PR #30, opened 2017-07-24
Use #include <bits/extc++.h> header where needed

Tyilo:


simonlindholm:
Didn't know this existed! Thanks!

PR #32, opened 2017-08-09
Use std::tie in operator< and ==

Tyilo:
Nice for `operator<` and easily copyable to implement `operator=`


simonlindholm:
Oh, huh, didn't know that std::tie yielded regular tuples. Angle::operator< doesn't compile in your version, because you can't take an (lvalue) reference to `quad()`, but the rest of the changes look fine.

Tyilo:
Fixed now :)

simonlindholm:
Thanks!

PR #34, opened 2017-09-01
Add the Z-function

jsannemo:
The Z-function computes the longest common prefix of the string itself.
Uses the pi array instead to be a bit shorter. Also, the conversion itself is good to document.

### DIFF
--- a/content/strings/KMP.h
+++ b/content/strings/KMP.h
@@ -3,10 +3,13 @@
  * License: CC0
  * Description: pi[x] computes the length of the longest prefix of s that ends at x, other than s[0..x] itself
  * This is used by find to find all occurances of a string.
+ * z[x] computes the longest prefix of s that starts at x. Note that z[0] = 0 though, which you often want.
+ * It takes pi as input
  * Time: O(pattern) for pi, O(word + pattern) for find
  * Status: Tested on Kattis, stringmatching
  * Usage:
  *  vi p = pi(pattern); vi occ = find(word, p);
+ *  vi zf = z(p);
  */
 #pragma once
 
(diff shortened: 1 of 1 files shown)


content/strings/KMP.h
```diff
@@ -20,6 +23,15 @@ vi pi(const string& s) {
 	return p;
 }
 
+vi z(const vector<int>& p) {
```
simonlindholm: `vi`

content/strings/KMP.h
```diff
@@ -3,10 +3,13 @@
  * License: CC0
  * Description: pi[x] computes the length of the longest prefix of s that ends at x, other than s[0..x] itself
  * This is used by find to find all occurances of a string.
+ * z[x] computes the longest prefix of s that starts at x. Note that z[0] = 0 though, which you often want.
+ * It takes pi as input
```
simonlindholm: add a trailing `.`

content/strings/KMP.h
```diff
@@ -20,6 +23,15 @@ vi pi(const string& s) {
 	return p;
 }
 
+vi z(const vector<int>& p) {
+	vi Z(sz(p));
+    	rep(i,0,sz(p))
+        	Z[i-p[i]+1] = max(Z[i-p[i]+1], p[i]);
+    	rep(i,0,sz(p)) if (p[i] && Z[p[i] - 1])
+        	Z[i] = max(Z[i], P[Z[P[i] - 1] - 1]);
```
simonlindholm: is that a capital `P`? (does this compile?)

simonlindholm:
Does KACTL still fit in 25 pages with that addition? If not, remove some `\hardcolumnbreak`, `\columnbreak`, `newpage` to make it fit (I added some before ICPC to make things look nicer).

content/strings/KMP.h
jsannemo: Darnit, I thought my coding-in-the-edit-field would work :(

simonlindholm: Now add a test for penance!

simonlindholm:
Tested it today, seems to be broken for `abacaba` (gives 0010300 instead of 0010301).

content/strings/KMP.h
ludopulles: See https://github.com/ludopulles/kactl/commit/e1fea999ec000dd677ec24515361abf844e9ec35#diff-4c06cbaa19a2ea174dc598afa2a9c903 for some  testcode :smile:.

simonlindholm:
Replaced by #66.

PR #36, opened 2017-09-20
Fix compiler warning in 2sat.h

Tyilo:
The expression `1 - x&1` produces a warning since the compiler suggests that we should add parentheses so it becomes `(1 - x)&1`.

In this specific case however, we have that `(1 - x)&1 == 1 - (x&1) == !(x&1)`.

I suggest we use `!(x&1)`, but we could also pick any of the two others with parentheses.

### DIFF
--- a/content/graph/2sat.h
+++ b/content/graph/2sat.h
@@ -59,7 +59,7 @@ struct TwoSat {
 			x = z.back(); z.pop_back();
 			comp[x] = time;
 			if (values[x>>1] == -1)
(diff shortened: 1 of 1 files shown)


simonlindholm:
Heh, that's funny. I guess I initially wrote it intending `1 - (x&1)`, and testing didn't find any errors. Your suggested fix sounds good.

Tyilo:
Yeah I guess that was the case, but after staring at it long enough I realized the placement of the parentheses didn't matter :)

PR #39, opened 2017-09-28
Use unordered_map for edmondsKarp

Tyilo:
It's a lot faster

### DIFF
--- a/content/graph/edmondsKarp.h
+++ b/content/graph/edmondsKarp.h
@@ -9,7 +9,7 @@
  */
 #pragma once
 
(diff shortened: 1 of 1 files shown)


simonlindholm:
How much faster? Is it the same for sparse graphs and dense graphs? It's not obvious to me how the performance of traversing a map vs. unordered_map compares.

Tyilo:
Testing on https://open.kattis.com/problems/maxflow I get `0.40s` CPU time when using `unordered_map` and `0.70s` when using `map`.

simonlindholm:
Seems to be faster for most graphs indeed (I did see some exceptions, but they don't seem to be the norm).

PR #40, opened 2017-10-05
typo in PolyRoots example

LudwikJaniuk:
polynomial -> Polynomial, as the class is called in Polynomial.h


simonlindholm:
Thanks!

PR #41, opened 2017-10-06
Fix indices in toposort

Tyilo:
Before this change they were 1-indexed.


simonlindholm:
Makes sense. (With "Source: predates tinyKACTL" I'm not surprised at the old behavior...)

PR #43, opened 2017-10-29
Fix factor.h

Tyilo:
`pr` shouldn't be local

### DIFF
--- a/content/number-theory/factor.h
+++ b/content/number-theory/factor.h
@@ -6,13 +6,14 @@
  * Description: Pollard's rho algorithm. It is a probabilistic factorisation
  * algorithm, whose expected time complexity is good. Before you start using it,
  * run {\tt init(bits)}, where bits is the length of the numbers you use.
+ * Returns factors of the input without duplicates.
  * Time: Expected running time should be good enough for 50-bit numbers.
  */
 #pragma once
 
+#include "ModMulLL.h"
 #include "MillerRabin.h"
 #include "eratosthenes.h"
-#include "euclid.h"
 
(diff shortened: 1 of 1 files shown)


simonlindholm:
You really seem to find all the crufty corners of KACTL... While you're changing factor.h, do you mind:

- using `pr.assign(all(p));` instead of copy(..., back_inserter(pr))
- using `__gcd` instead of `gcd`, and removing the include of `euclid.h`
- adding an include for `ModMulLL.h`
- adding a note to the documentation that it returns the factors of the number *without duplicates*

?

Tyilo:
Do we want to include `ModMulLL.h` when `MillerRabin.h` includes `ModMullLL.h`?

Shouldn't there be a relation between `const init bits` in `ModMulLL.h` and `void init(int bits)` in `factor.h`?
Maybe replace `init` with:

```
void init() {
  vi p = eratosthenes_sieve(1 << ((64 - bits + 2) / 3));
  pr.assign(all(p));
}
```

simonlindholm:
> Do we want to include ModMulLL.h when MillerRabin.h includes ModMullLL.h?

Ah, I missed that. I guess it still doesn't hurt?

> Shouldn't there be a relation between const init bits in ModMulLL.h and void init(int bits) in factor.h?

Mmm, perhaps. No real opinion, honestly -- I can see arguments both ways, and I haven't actually ever used this algorithm in a contest, so I don't care that much about it.

Tyilo:
Done :)

simonlindholm:
Thanks!

PR #53, opened 2018-10-23
Add support for python files with descriptions

Tyilo:
Supports the syntax:

```
'''
 * Author: Asger Hautop Drewsen
 * License: CC0
 * Description: Prints Foo
 * Status: Works
 * Time: $O(1)$
'''

print('Foo')
```


build/kactlprocessor.py
```diff
@@ -84,18 +99,18 @@ def processwithcomments(caption, instream, outstream, listingslang = None):
...
     end = 0
     commands = {}
     while start >= 0 and not error:
         nsource = nsource.rstrip() + source[end:start]
-        end = source.find("*/", start)
+        end, endlen = find_end_comment(source, start + 1)
         if end<start:
             error = "Invalid /** */ comments."
             break
         comment = source[start+3:end].strip()
```
simonlindholm: so that this `start+3` becomes ... `start_end`? `start2`? something like that

build/kactlprocessor.py
```diff
@@ -52,6 +52,21 @@ def addref(caption, outstream):
...
+        if i != -1 and (i < first[0] or first[0] == -1):
+            first = (i, len(s))
+
+    return first
+
+def find_start_comment(source, start=None):
+    return find_helper(source, start, ['/**', "'''"])
+
+def find_end_comment(source, start=None):
+    return find_helper(source, start, ['*/', "'''"])
```
simonlindholm: is it worth supporting `"""`? should we try to pair comment end syntax with the matching start syntax?

Tyilo:
@simonlindholm Should be better now.
It would be better if it was language-specific, but I think it's good enough for now.

simonlindholm:
Very nice!

PR #56, opened 2019-03-29
Changed Miller Rabin to deterministic variant

Chillee:
It's about 2x faster for <1e9 and 3x faster for ~1e15. 

Also, the determinism is nice. Currently, the probabilistic Miller Rabin with 15 iterations is expected to fail in about a thousand seconds worth of computation on worst case input.

### DIFF
--- a/content/number-theory/MillerRabin.h
+++ b/content/number-theory/MillerRabin.h
@@ -20,7 +20,7 @@ bool isPrime(ull n) {
         ull p = mod_pow(a, d, n), i = d;
         while (p != n - 1 && i != n - 1 && p != 1)
             i <<= 1, p = mod_mul(p, p, n);
-        if (i == n - 1 && i > d) return false;
+        if (p != n-1 && i != d) return false;
     }
     return true;
 }
--- a/content/number-theory/MillerRabin.h
+++ b/content/number-theory/MillerRabin.h
@@ -1,10 +1,10 @@
 /**
- * Author: chilli, c1729
+ * Author: chilli, c1729, Simon Lindholm
  * Date: 2019-03-28
  * License: CC0
- * Source: Bases taken from https://miller-rabin.appspot.com/
- * Description: Deterministic Miller-Rabin primality test. Guaranteed to work for numbers up to 2^64.
- * Can test 15 random numbers for $a$ for probabilistic test on values higher than 2^64.
+ * Source: Wikipedia, https://miller-rabin.appspot.com/
+ * Description: Deterministic Miller-Rabin primality test.
+ * Guaranteed to work for numbers up to 2^64; for larger numbers, extend A randomly.
  * Time: 7 times the complexity of $a^b \mod c$.
  */
 #pragma once
@@ -12,16 +12,14 @@
 #include "ModMulLL.h"
 
 bool isPrime(ull n) {
-    if (n <= 1) return false;
-    for (ull p : {2, 3, 5, 13, 19, 73, 193, 407521, 299210837})
-        if (n % p == 0) return (n == p);
-    ull d = n - 1;
-    while (!(d & 1)) d >>= 1;
-    for (ull a : {2, 325, 9375, 28178, 450775, 9780504, 1795265022}) { 
-        ull p = mod_pow(a, d, n), i = d;
-        while (p != n - 1 && i != n - 1 && p != 1)
-            i <<= 1, p = mod_mul(p, p, n);
-        if (p != n-1 && i != d) return false;
-    }
-    return true;
+	if (n < 2 || n % 2 == 0 || n % 3 == 0) return n - 2 < 2;
+	ull A[] = {2, 325, 9375, 28178, 450775, 9780504, 1795265022},
+	    s = __builtin_ctzll(n-1), d = n >> s;
+	trav(a, A) {   // ^ count trailing zeroes
+		ull p = mod_pow(a, d, n), i = s;
+		while (p != 1 && p != n - 1 && a % n && i--)
+			p = mod_mul(p, p, n);
+		if (p != n-1 && i != s) return 0;
+	}
+	return 1;
 }
--- a/content/number-theory/MillerRabin.h
+++ b/content/number-theory/MillerRabin.h
@@ -2,6 +2,7 @@
  * Author: chilli, c1729
  * Date: 2019-03-28
  * License: CC0
+ * Source: Bases taken from https://miller-rabin.appspot.com/
  * Description: Deterministic Miller-Rabin primality test. Guaranteed to work for numbers up to 2^64.
  * Can test 15 random numbers for $a$ for probabilistic test on values higher than 2^64.
  * Time: 7 times the complexity of $a^b \mod c$.
--- a/content/number-theory/MillerRabin.h
+++ b/content/number-theory/MillerRabin.h
@@ -2,9 +2,7 @@
  * Author: chilli, c1729
  * Date: 2019-03-28
  * License: CC0
- * Description: Deterministic Miller-Rabin primality test.
- * Guaranteed to be correct, 2x as probabilistic version for <1e9, 3x as fast
- * for 1e15
+ * Description: Deterministic Miller-Rabin primality test. Numbers taken from 
  * Time: 7 times the complexity of $a^b \mod c$.
  */
 #pragma once
@@ -12,10 +10,10 @@
 #include "ModMulLL.h"
 
 bool prime(ull n) {
-    vector<ull> ps({2, 3, 5, 13, 19, 73, 193, 407521, 299210837});
     vector<ull> cs({2, 325, 9375, 28178, 450775, 9780504, 1795265022});
-    if (n <= 1 || any_of(all(ps), [&](ull p) { return n % p == 0; }))
-        return count(all(ps), n) > 0;
+    if (n <= 1) return false;
+    for (ull p : {2, 3, 5, 13, 19, 73, 193, 407521, 299210837})
+        if (n % p == 0) return (n == p);
     ull d = n - 1, s = 0;
     while (!(d & 1)) d >>= 1, s++;
     return !any_of(all(cs), [&](ull a) {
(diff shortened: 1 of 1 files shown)


content/number-theory/MillerRabin.h
```diff
@@ -1,30 +1,27 @@
...
-	}
-	return true;
+bool prime(ull n) {
+    vector<ull> ps({2, 3, 5, 13, 19, 73, 193, 407521, 299210837});
+    vector<ull> cs({2, 325, 9375, 28178, 450775, 9780504, 1795265022});
+    if (n <= 1 || any_of(all(ps), [&](ull p) { return n % p == 0; }))
+        return count(all(ps), n) > 0;
+    ull d = n - 1, s = 0;
+    while (!(d & 1)) d >>= 1, s++;
+    return !any_of(all(cs), [&](ull a) {
```
simonlindholm: The `any_of` involves quite a bit of boilerplate and perhaps overhead if the compiler fails to inline it. How about
```c++
for (ull a : {2, 325, 9375, 28178, 450775, 9780504, 1795265022})
  for (ull i = 0; ...) ...
return true;

File: content/number-theory/MillerRabin.h [hunk 2]
```diff
@@ -1,30 +1,27 @@
...
-			tmp *= 2;
-		}
-		if (mod != p - 1 && tmp % 2 == 0) return false;
-	}
-	return true;
+bool prime(ull n) {
+    vector<ull> ps({2, 3, 5, 13, 19, 73, 193, 407521, 299210837});
+    vector<ull> cs({2, 325, 9375, 28178, 450775, 9780504, 1795265022});
+    if (n <= 1 || any_of(all(ps), [&](ull p) { return n % p == 0; }))
+        return count(all(ps), n) > 0;
```
simonlindholm: this looks a bit awkward to type, I think I'd prefer making it
```c++
if (n <= 1) return 0;
for (int p : {2, 3, 5, 13, 19, 73, 193, 407521, 299210837})
  if (n % p == 0) return (n == p);
```

content/number-theory/MillerRabin.h
```diff
@@ -1,30 +1,27 @@
...
+ * Date: 2019-03-28
  * License: CC0
- * Source: TopCoder tutorial
- * Description: Miller-Rabin primality probabilistic test.
- * Probability of failing one iteration is at most 1/4. 15 iterations should be
- * enough for 50-bit numbers.
- * Time: 15 times the complexity of $a^b \mod c$.
+ * Description: Deterministic Miller-Rabin primality test.
+ * Guaranteed to be correct, 2x as probabilistic version for <1e9, 3x as fast
+ * for 1e15
```
simonlindholm: Remove the mention of the probabilistic version

content/number-theory/MillerRabin.h
```diff
@@ -1,30 +1,27 @@
...
-		while (tmp != p - 1 && mod != 1 && mod != p - 1) {
-			mod = mod_mul(mod, mod, p);
-			tmp *= 2;
-		}
-		if (mod != p - 1 && tmp % 2 == 0) return false;
-	}
-	return true;
+bool prime(ull n) {
+    vector<ull> ps({2, 3, 5, 13, 19, 73, 193, 407521, 299210837});
+    vector<ull> cs({2, 325, 9375, 28178, 450775, 9780504, 1795265022});
```
simonlindholm: Can you add a reference for these numbers? For which range of `n` do they suffice? I can imagine wanting to reimplement this in Python with bignums during a contest, so 2**64 isn't always enough and we should describe how to safely extend it to larger n (is "test 15 a's at random" the right method for that?).

cheran-senthil: Well, it was originally some Python code I made, from [here](https://github.com/cheran-senthil/PyRival/blob/master/snippets/Mathematics/is_prime.py), perhaps you can come up with a cleaner convertion to C++. You can safely extend to more numbers if you are checking for over 2**64, by taking [more smaller primes](https://en.wikipedia.org/wiki/Miller%E2%80%93Rabin_primality_test#Deterministic_variants) upto 2(ln n)2

Chillee: I think if you want to change it to a randomized version that'll "work" for arbitrarily large numbers, you can just select 15 random a's.

You shouldn't need to change anything else.

content/number-theory/MillerRabin.h
Chillee: The semantics here are a bit annoying to duplicate with regular loops - largely because you need to break out of two nested loops.

Here's a couple of other options I came up with, in varying levels of jankness/seriousness

```cpp
for (ull a : {2, 325, 9375, 28178, 450775, 9780504, 1795265022}) {
    for (ull i = 0, p = binExp(a, d, n); i < s; i++, p = binExp(p, 2, n))
        if (p == n - 1 || (i == 0 && p == 1))
            goto end;
    return false;
end:;
}
```

```cpp
for (ull a : {2, 325, 9375, 28178, 450775, 9780504, 1795265022}) {
    ull p = binExp(a, d, n), i = 0;
    if (p == 1)
        continue;
    for (; p != n - 1 && i < s; i++)
        p = binExp(p, 2, n);
    if (i == s)
        return false;
}
return true;
```

```cpp
for (ull a : {2, 325, 9375, 28178, 450775, 9780504, 1795265022}) {
    if ([&] {
            for (ull i = 0, p = binExp(a, d, n); i < s; i++, p = binExp(p, 2, n))
                if (p == n - 1 || (i == 0 && p == 1))
                    return false;
            return true;
        }())
        return false;
}
return true;
```
```cpp
for (ull a : {2, 325, 9375, 28178, 450775, 9780504, 1795265022}) {
    bool found = false;
    for (ull i = 0, p = binExp(a, d, n); i < s; i++, p = binExp(p, 2, n))
        if (p == n - 1 || (i == 0 && p == 1)) {
            found = true;
            break;
        }
    if (!found)
        return false;
}
return true;
```

content/number-theory/MillerRabin.h
Chillee: Also, to be clear, the numbers are taken from https://miller-rabin.appspot.com/

content/number-theory/MillerRabin.h
Chillee: I like the last one the most

content/number-theory/MillerRabin.h
Chillee: Also, where should I add this reference? Seems unnecessary to put the source of the numbers in the notebook itself, but I'm not sure what's appropriate. Perhaps the source/authors?

content/number-theory/MillerRabin.h
simonlindholm: Ah, I see. Agreed, the last one looks good. Curiously, it's very similar to KACTL's current version, except that the last condition is `if (i == n - 1 && i > d) return false;` rather than `if (mod != p - 1 && tmp % 2 == 0) return false;`. I don't know what that difference means...

content/number-theory/MillerRabin.h
simonlindholm: Source would be good.

content/number-theory/MillerRabin.h
Chillee: Thinking about it more... I think what I have right now has a slight bug. Gimme a bit to fix it.

Chillee: The conditions that are equivalent (and this isn't obvious, it relies on specific knowledge about Miller-Rabin) are `i==n-1 && p!=n-1` or `p!=n-1 && i!=d` (the second one is more directly equivalent to the current KACTL condition).

I'll stick with the current KACTL condition - it's shorter :^)

Chillee:
I think all issues have been addressed.

simonlindholm:
I found some time to look into this. How do you feel about the following shortening?

```cpp
bool isPrime(ull n) {
	if (n <= 1 || n % 2 == 0) return n == 2;
	ull A[] = {2, 325, 9375, 28178, 450775, 9780504, 1795265022},
	    s = __builtin_ctzll(n-1), d = (n-1) >> s;
	trav(a, A) {   // ^ count trailing zeroes
		ull p = mod_pow(a, d, n), i = s;
		while (p != 1 && p != n - 1 && a % n && i--)
			p = mod_mul(p, p, n);
		if (p != n-1 && i != s) return 0;
	}
	return 1;
}
```

Changes being:
- put the array A on its own line to comply with the 63-char limit
- compute the logarithm of `(p-1)/s` using `__builtin_ctzll`, and use that as a loop counter instead of i going from `(p-1)/s` and being multiplied by 2 in each step. It's slightly shorter and gets rid of a comma operator in the while loop in favor of `i--` in the conditional.
  * add a comment for what __builtin_ctzll does in case you need to reimplement the algorithm in e.g. Python.
- get rid of the initial divisibility checks for factors of the bases, except for 2, and skip the inner loop if `a % n == 0`. Reasoning:
  * this only makes the a difference in the case where n shares a factor with one of the numbers in A
  * if `a % n == 0`, n is one of the divisors of the numbers in A. I checked them all and isPrime gave the correct result
  * otherwise, it's harmless to do the test, and if there is actually a shared factor, we will always return "composite" because raising `a` to some power cannot get rid of that factor, so we can never reach `n-1` or `1`.

I double-checked that it works up to n = 5e7.

Chillee:
I mean, it mostly looks pretty good to me.

It is possible to recover most of the performance back just by changing the first line to
```
	if (n <= 1 || n % 2 == 0 || n%3 == 0) return n == 2 || n==3;
```

Not sure if that's worth it - I'll leave that up to your judgement.

The performance disparity also gets worse as the starting value increases. For example, starting from `1e14` and testing the `1e6` values after that, the new version appears to run about 50% slower (with the minor change it's about 25% slower).

Once again, not sure how you feel about the importance of performance vs code-size here - I haven't used Miller Rabin enough in contest to evaluate the tradeoffs meaningfully. I'll defer to your judgement here.

simonlindholm:
> I haven't used Miller Rabin enough in contest to evaluate the tradeoffs meaningfully.

Me neither. I guess it doesn't hurt, though, since it's such a small change (`	if (n < 2 || n % 2 == 0 || n % 3 == 0) return n - 2 < 2;` if you minify it a bit). I'll merge with that.

PR #57, opened 2019-04-05
bugfix subtraction NTT

Noobgam:
This has to be a bug, right?

I'd rather use functions of sorts `modadd/modsub` to avoid such mistakes, and avoid multiple modulo operations (1 for t instead of 3 in this case)


simonlindholm:
No, this is intended. In my experience it's usually best to reduce the answer to [0, mod-1] only when printing or when comparing values for equality -- reducing to [0, mod-1] for every subtraction is verbose, slow and error-prone.

However, it might be reasonable to loop over all entries of the returned vector at the very end of the NTT and reduce them (`trav(x, ret) if (x < 0) x += mod;`), in case people don't expect this. That would preserve performance.

Noobgam:
Got it, didn't quite notice that in documentation. Fine with me.

PR #58, opened 2019-04-11
Replaced with new suffix array implementation

Chillee:
Should be around 10-15x faster than the current KACTL implementation for N=1e6+5.

Benchmarks here: https://ideone.com/GpSu2m (there's a small bug in the implementation used in this benchmark that's corrected in the PR - `wa` and `wb` need to be double the size of the string in order to not cause issues.

Correctness test here: https://ideone.com/oeR3jz


simonlindholm:
Thanks, this is excellent! I updated your PR with some formatting changes, as well as changing wa and wb back to have size n because I couldn't find any issues with that with ASAN. Where did it blow up for you? I also wrote some comprehensive bruteforce tests.

"License: unknown" is sad, though :/

PR #59, opened 2019-04-16
Updated NTT to be similar to the rewritten FFT

Chillee:
Since we rewrote the FFT and got a massive speedup, seems strange that the NTT has such a different implementation from the FFT.

This PR updates the NTT to be written in the style of the new FFT, and gets a 7-10x boost in performance from doing so: https://ideone.com/QJwGfE

This also includes an update to the fuzz test for ntt - A. because the new FFT implementation returns empty vector if either one of the input polynomials is empty, and that's the API I copied for the NTT, and B. because the new NTT now requires that all inputs be within the mod range.

This is still somewhat WIP - I have a couple questions/issues I need to investigate more (and some notes)
1. The conditional modular adds/subtracts add about 30% in performance UNLESS `O3` is enabled. If `O3` is enabled then it slows it down by about 50%. In this case, `O3` is harming the performance of the conditional modular adds/subtracts. It goes from about 200 ms to 450 ms with `O3`. I don't understand why, and I can't replicate it in an isolated example.

2. Is it possible to shorten this:
```cpp
    int curL = (mod - 1) >> 2;
    for (int k = 2; k < n; k *= 2) {
        int z[] = {1, modpow(root, curL)};
        curL >>= 1;
        rep(i, k, 2 * k) rt[i] = mul(rt[i / 2], z[i & 1]);
    }
```
3. Handling the output vector. It's currently handled in a redundant way, and I feel like it can be cleaned up nicely.

4. Re: the modular operations, it may be a good idea to inline them into the code (especially add/subtract since they're only used once).

I also need to test it more thoroughly, I'm not absolutely confident there's no issues with certain sizes and what not.

### DIFF
--- a/content/numerical/NumberTheoreticTransform.h
+++ b/content/numerical/NumberTheoreticTransform.h
@@ -36,13 +36,12 @@ vl conv(const vl& a, const vl& b) {
 	vl L(a), R(b), out(n), rt(n, 1), rev(n);
 	L.resize(n), R.resize(n);
 	rep(i,0,n) rev[i] = (rev[i / 2] | (i & 1) << B) / 2;
-	int curL = mod / 2;
+	ll curL = mod / 2, inv = modpow(n, mod - 2);
 	for (int k = 2; k < n; k *= 2) {
 		ll z[] = {1, modpow(root, curL /= 2)};
 		rep(i,k,2*k) rt[i] = rt[i / 2] * z[i & 1] % mod;
 	}
-	ntt(L, rt, rev, n), ntt(R, rt, rev, n);
-	ll inv = modpow(n, mod - 2);
+	ntt(L, rt, rev, n); ntt(R, rt, rev, n);
 	rep(i,0,n) out[-i & (n-1)] = L[i] * R[i] % mod * inv % mod;
 	ntt(out, rt, rev, n);
 	return {out.begin(), out.begin() + s};
--- a/content/numerical/NumberTheoreticTransform.h
+++ b/content/numerical/NumberTheoreticTransform.h
@@ -19,32 +19,31 @@ const ll mod = (119 << 23) + 1, root = 62; // = 998244353
 // and 483 << 21 (same root). The last two are > 10^9.
 
 typedef vector<ll> vl;
-void ntt(vl &a, vl &rt, vl &rev, int n) {
+void ntt(vl& a, vl& rt, vl& rev, int n) {
 	rep(i,0,n) if (i < rev[i]) swap(a[i], a[rev[i]]);
 	for (int k = 1; k < n; k *= 2)
-		for (int i = 0; i < n; i += 2 * k)
-			rep(j,0,k) {
+		for (int i = 0; i < n; i += 2 * k) rep(j,0,k) {
 				ll z = rt[j + k] * a[i + j + k] % mod, &ai = a[i + j];
-				a[i + j + k] = z > ai ? ai - z + mod : ai - z;
-				ai += ai + z >= mod ? z - mod : z;
-			}
+				a[i + j + k] = (z > ai ? ai - z + mod : ai - z);
+				ai += (ai + z >= mod ? z - mod : z);
+	}
 }
 
-vl conv(const vl &a, const vl &b) {
+vl conv(const vl& a, const vl& b) {
 	if (a.empty() || b.empty())
 		return {};
-	int s = sz(a) + sz(b) - 1, B = 32 - __builtin_clz(s), n = 1 << B;
+	int s = sz(a)+sz(b)-1, B = 32 - __builtin_clz(s), n = 1 << B;
 	vl L(a), R(b), out(n), rt(n, 1), rev(n);
 	L.resize(n), R.resize(n);
 	rep(i,0,n) rev[i] = (rev[i / 2] | (i & 1) << B) / 2;
 	int curL = mod / 2;
 	for (int k = 2; k < n; k *= 2) {
 		ll z[] = {1, modpow(root, curL /= 2)};
-		rep(i,k,2*k) rt[i] = rt[i / 2] * z[i & 1 ] %mod;
+		rep(i,k,2*k) rt[i] = rt[i / 2] * z[i & 1] % mod;
 	}
 	ntt(L, rt, rev, n), ntt(R, rt, rev, n);
 	ll inv = modpow(n, mod - 2);
-	rep(i,0,n) out[-i&(n-1)] = L[i] * R[i] % mod * inv % mod;
+	rep(i,0,n) out[-i & (n-1)] = L[i] * R[i] % mod * inv % mod;
 	ntt(out, rt, rev, n);
 	return {out.begin(), out.begin() + s};
 }
--- a/content/numerical/NumberTheoreticTransform.h
+++ b/content/numerical/NumberTheoreticTransform.h
@@ -33,10 +33,10 @@ void ntt(vl &a, vl &rt, vl &rev, int n) {
 vl conv(const vl &a, const vl &b) {
 	if (a.empty() || b.empty())
 		return {};
-	int s = sz(a) + sz(b) - 1, L = 32 - __builtin_clz(s), n = 1 << L;
+	int s = sz(a) + sz(b) - 1, B = 32 - __builtin_clz(s), n = 1 << B;
 	vl L(a), R(b), out(n), rt(n, 1), rev(n);
 	L.resize(n), R.resize(n);
-	rep(i,0,n) rev[i] = (rev[i / 2] | (i & 1) << L) / 2;
+	rep(i,0,n) rev[i] = (rev[i / 2] | (i & 1) << B) / 2;
 	int curL = mod / 2;
 	for (int k = 2; k < n; k *= 2) {
 		ll z[] = {1, modpow(root, curL /= 2)};
--- a/content/numerical/NumberTheoreticTransform.h
+++ b/content/numerical/NumberTheoreticTransform.h
@@ -14,9 +14,9 @@
 
 #include "../number-theory/ModPow.h"
 
-const ll mod = (119 << 23) + 1, root = 3; // = 998244353
-// For p < 2^30 there is also e.g. (5 << 25, 3), (7 << 26, 3),
-// (479 << 21, 3) and (483 << 21, 5). The last two are > 10^9.
+const ll mod = (119 << 23) + 1, root = 62; // = 998244353
+// For p < 2^30 there is also e.g. 5 << 25, 7 << 26, 479 << 21
+// and 483 << 21 (same root). The last two are > 10^9.
 
 typedef vector<ll> vl;
 void ntt(vl &a, vl &rt, vl &rev, int n) {
--- a/content/numerical/NumberTheoreticTransform.h
+++ b/content/numerical/NumberTheoreticTransform.h
@@ -1,12 +1,12 @@
 /**
- * Author: Simon Lindholm
- * Date: 2016-09-10
+ * Author: chilli
+ * Date: 2019-04-16
  * License: CC0
  * Source: based on KACTL's FFT
  * Description: Can be used for convolutions modulo specific nice primes
  * of the form $2^a b+1$, where the convolution result has size at most $2^a$.
- * For other primes/integers, use two different primes and combine with CRT.
- * Inputs must be in [0, mod)
+ * For other primes/integers, use three different primes and combine with CRT.
+ * Inputs must be in [0, mod).
  * Time: O(N \log N)
  * Status: Somewhat tested
  */
--- a/content/numerical/NumberTheoreticTransform.h
+++ b/content/numerical/NumberTheoreticTransform.h
@@ -6,49 +6,45 @@
  * Description: Can be used for convolutions modulo specific nice primes
  * of the form $2^a b+1$, where the convolution result has size at most $2^a$.
  * For other primes/integers, use two different primes and combine with CRT.
- * May return negative values.
+ * Inputs must be in [0, mod)
  * Time: O(N \log N)
  * Status: Somewhat tested
  */
 #pragma once
 
-#include "ModPow.h"
+#include "../number-theory/ModPow.h"
 
 const ll mod = (119 << 23) + 1, root = 3; // = 998244353
 // For p < 2^30 there is also e.g. (5 << 25, 3), (7 << 26, 3),
 // (479 << 21, 3) and (483 << 21, 5). The last two are > 10^9.
 
 typedef vector<ll> vl;
-void ntt(ll* x, ll* temp, ll* roots, int N, int skip) {
-	if (N == 1) return;
-	int n2 = N/2;
-	ntt(x     , temp, roots, n2, skip*2);
-	ntt(x+skip, temp, roots, n2, skip*2);
-	rep(i,0,N) temp[i] = x[i*skip];
-	rep(i,0,n2) {
-		ll s = temp[2*i], t = temp[2*i+1] * roots[skip*i];
-		x[skip*i] = (s + t) % mod; x[skip*(i+n2)] = (s - t) % mod;
-	}
+void ntt(vl &a, vl &rt, vl &rev, int n) {
+	rep(i,0,n) if (i < rev[i]) swap(a[i], a[rev[i]]);
+	for (int k = 1; k < n; k *= 2)
+		for (int i = 0; i < n; i += 2 * k)
+			rep(j,0,k) {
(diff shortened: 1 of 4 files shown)


content/numerical/NumberTheoreticTransform.h
```diff
@@ -18,37 +18,41 @@ const ll mod = (119 << 23) + 1, root = 3; // = 998244353
...
+    int curL = (mod - 1) >> 2;
+    for (int k = 2; k < n; k *= 2) {
+        int z[] = {1, modpow(root, curL)};
+        curL >>= 1;
+        rep(i, k, 2 * k) rt[i] = mul(rt[i / 2], z[i & 1]);
+    }
+    copy(all(a), begin(inl)), copy(all(b), begin(inr));
+    ntt(inl, rt, rev, n), ntt(inr, rt, rev, n);
+    int invN = modpow(n, mod - 2);
+    rep(i, 0, n) out[i] = mul(mul(inl[i], inr[i]), invN);
```
simonlindholm: `out[-i & (n-1)] = ...` to avoid the reverse

content/numerical/NumberTheoreticTransform.h
```diff
@@ -18,37 +18,41 @@ const ll mod = (119 << 23) + 1, root = 3; // = 998244353
...
+    reverse(out.begin() + 1, out.begin() + n);
+    ntt(out, rt, rev, n);
+    rep(i, 0, sz(res)) res[i] = out[i];
+    return out;
```
simonlindholm: `return res;`, or `return {out.begin(), out.end() + s};` for some `s` declared higher up (the FFT needs to convert complex->real, but here things are simpler)

fuzz-tests/ntt.cpp
```diff
@@ -21,49 +21,52 @@ ll modpow(ll a, ll e) {
...
+vl conv(const vi &a, const vi &b) {
+    if (a.empty() || b.empty())
+        return {};
+    vl res(sz(a) + sz(b) - 1);
+    int L = 32 - __builtin_clz(sz(res)), n = 1 << L;
+    vi inl(n), inr(n);
+    vi out(n), rt(n, 1);
+    vi rev(n);
+    rep(i, 0, n) rev[i] = (rev[i / 2] | (i & 1) << L) / 2;
+    int curL = (mod - 1) >> 2;
```
simonlindholm: `mod / 4`

fuzz-tests/ntt.cpp
```diff
@@ -21,49 +21,52 @@ ll modpow(ll a, ll e) {
...
+    for (int k = 2; k < n; k *= 2) {
+        int z[] = {1, modpow(root, curL)};
```
simonlindholm: `curL /= 2` as argument (and use `/ 2` above)?

or `curL /= 2` below; it shaves off one character and compiles the same

fuzz-tests/ntt.cpp
```diff
@@ -21,49 +21,52 @@ ll modpow(ll a, ll e) {
...
+        curL >>= 1;
+        rep(i, k, 2 * k) rt[i] = mul(rt[i / 2], z[i & 1]);
+    }
+    copy(all(a), begin(inl)), copy(all(b), begin(inr));
```
simonlindholm: an alternative to this would be
```
vi inl(all(a)), inr(all(b));
inl.resize(n); inr.resize(n);
```
(it adds some memory allocations but hopefully it's not that much slower). OTOH that might not be a LOC win when putting all `vi` declarations on the same line.

fuzz-tests/ntt.cpp
```diff
@@ -21,49 +21,52 @@ ll modpow(ll a, ll e) {
...
-	return c;
+
+vl conv(const vi &a, const vi &b) {
+    if (a.empty() || b.empty())
+        return {};
+    vl res(sz(a) + sz(b) - 1);
+    int L = 32 - __builtin_clz(sz(res)), n = 1 << L;
+    vi inl(n), inr(n);
+    vi out(n), rt(n, 1);
+    vi rev(n);
```
simonlindholm: same line for all of these

fuzz-tests/ntt.cpp
```diff
@@ -21,49 +21,52 @@ ll modpow(ll a, ll e) {
...
+int sub(int a, int b) { return b > a ? a - b + mod : a - b; }
+int add(int a, int b) { return a + b >= mod ? a + b - mod : a + b; }
+void ntt(vi &a, vi &rt, vi &rev, int n) {
+    rep(i, 0, n) if (i < rev[i]) swap(a[i], a[rev[i]]);
+    for (int k = 1; k < n; k *= 2)
+        for (int i = 0; i < n; i += 2 * k)
+            rep(j, 0, k) {
+                int z = mul(rt[j + k], a[i + j + k]);
+                a[i + j + k] = sub(a[i + j], z);
+                a[i + j] = add(a[i + j], z);
```
simonlindholm: Did this improve performance much over the previous version? This requires inputs to be within [0, mod), and adds add/sub functions.

fuzz-tests/ntt.cpp
```diff
@@ -21,49 +21,52 @@ ll modpow(ll a, ll e) {
...
-		x[skip*i] = (s + t) % mod; x[skip*(i+n2)] = (s - t) % mod;
-	}
-}
-void ntt(vl& x, bool inv = false) {
-	ll e = modpow(root, (mod-1) / sz(x));
-	if (inv) e = modpow(e, mod-2);
-	vl roots(sz(x), 1), temp = roots;
-	rep(i,1,sz(x)) roots[i] = roots[i-1] * e % mod;
-	ntt(&x[0], &temp[0], &roots[0], sz(x), 1);
+int mul(int a, int b) { return (ll)a * b % mod; }
```
simonlindholm: You get a -Wconversion warning here, and in a bunch of other places. -Wconversion is used by KACTL's default compile macro, and super important for not getting overflows in number theory-heavy code, so please make sure the code compiles without such warnings. Possibly by using `ll` for all modded numbers (`int` and `ll` should be about as fast on 64-bit systems, and even on 32-bit ones I think perf might be dominated by modulo operations? So with some luck this won't slow anything down. But please do benchmark...)

fuzz-tests/ntt.cpp
```diff
@@ -21,49 +21,52 @@ ll modpow(ll a, ll e) {
...
+    int curL = (mod - 1) >> 2;
+    for (int k = 2; k < n; k *= 2) {
+        int z[] = {1, modpow(root, curL)};
+        curL >>= 1;
+        rep(i, k, 2 * k) rt[i] = mul(rt[i / 2], z[i & 1]);
+    }
+    copy(all(a), begin(inl)), copy(all(b), begin(inr));
+    ntt(inl, rt, rev, n), ntt(inr, rt, rev, n);
+    int invN = modpow(n, mod - 2);
+    rep(i, 0, n) out[i] = mul(mul(inl[i], inr[i]), invN);
```
simonlindholm: can we reuse `inl` instead of having a new `out` vector?

fuzz-tests/ntt.cpp
```diff
@@ -21,49 +21,52 @@ ll modpow(ll a, ll e) {
 }
 
 typedef vector<ll> vl;
```
simonlindholm: Ooops, I started commenting on the fuzz test rather than the original code. Is it possible to use an `#include` here? Perhaps with some trickery for getting the right `mod`, e.g. `namespace ignored { #include "modpow.h" } #include "ntt.h"` or `#define`s

simonlindholm:
> This PR updates the NTT to be written in the style of the new FFT, and gets a 7-10x boost in performance from doing so: https://ideone.com/QJwGfE

That's very nice! I thought mod operations would dominate perf and make this a smaller win, but apparently not. Is that number for 64-bit?

> The conditional modular adds/subtracts add about 30% in performance UNLESS O3 is enabled. If O3 is enabled then it slows it down by about 50%. In this case, O3 is harming the performance of the conditional modular adds/subtracts. It goes from about 200 ms to 450 ms with O3. I don't understand why, and I can't replicate it in an isolated example.

-O3.  ¯\\\_(ツ)\_/¯ I wouldn't worry so much about it, but checking the assembly might be interesting (`gcc -S`).

> Re: the modular operations, it may be a good idea to inline them into the code (especially add/subtract since they're only used once).

Yes please!

fuzz-tests/ntt.cpp
Chillee: You can also just do `vi inl(a), inr(b)`, right? so it's also a character win, at least.

fuzz-tests/ntt.cpp
Chillee: I don't easily see how if we're doing the `[-i & (n-i)]` for the bit reversal - perhaps you know how though.

simonlindholm: oh, good point, pick your poison :)

fuzz-tests/ntt.cpp
Chillee: https://ideone.com/emwmZA

About 25% gain. We can inline the add/sub functions.

Chillee:
Responded to all comments.

content/numerical/NumberTheoreticTransform.h
```diff
@@ -18,41 +18,33 @@ const ll mod = (119 << 23) + 1, root = 3; // = 998244353
...
     rep(i, 0, n) rev[i] = (rev[i / 2] | (i & 1) << L) / 2;
-    int curL = (mod - 1) >> 2;
+    int curL = (mod - 1) / 2;
     for (int k = 2; k < n; k *= 2) {
-        int z[] = {1, modpow(root, curL)};
-        curL >>= 1;
-        rep(i, k, 2 * k) rt[i] = mul(rt[i / 2], z[i & 1]);
+        ll z[] = {1, modpow(root, curL/=2)};
+        rep(i, k, 2 * k) rt[i] = rt[i / 2]* z[i & 1]%mod;
     }
```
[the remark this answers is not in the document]
Chillee: The roots are calculated in this way in FFT not only for numerical precision, but also for cache efficiency.

Chillee:
I think all review comments have been addressed on this.

content/numerical/NumberTheoreticTransform.h
```diff
@@ -18,41 +18,33 @@ const ll mod = (119 << 23) + 1, root = 3; // = 998244353
...
     rep(i, 0, n) if (i < rev[i]) swap(a[i], a[rev[i]]);
     for (int k = 1; k < n; k *= 2)
         for (int i = 0; i < n; i += 2 * k)
             rep(j, 0, k) {
-                int z = mul(rt[j + k], a[i + j + k]);
-                a[i + j + k] = sub(a[i + j], z);
-                a[i + j] = add(a[i + j], z);
+                ll z = rt[j + k] * a[i + j + k] % mod;
+                a[i + j + k] = z > a[i + j] ? a[i + j] - z + mod : a[i + j] - z;
+                a[i + j] = a[i + j] + z >= mod ? a[i + j] + z - mod : a[i + j] + z;
```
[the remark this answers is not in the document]
simonlindholm: I don't see this resolved -- we should definitely do it since this is over the character limit. Did you forget to push some commit?

simonlindholm:
See individual comment. Another thing to do before merging is to convert to tabs rather than spaces -- with tabs = 2 spaces in the pdf, 4-space indentation would be inconsistent. (Everyone misses this, so maybe we should just convert the whole KACTL repo over to spaces... but that's for another PR.)

content/numerical/NumberTheoreticTransform.h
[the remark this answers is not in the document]
Chillee: Whups... :sweat_smile:

Chillee:
Whups, you're right - didn't push. My bad.

content/numerical/NumberTheoreticTransform.h
```diff
@@ -12,43 +12,39 @@
...
+		rep(i,k,2*k) rt[i] = rt[i / 2]* z[i & 1]%mod;
 	}
-	a.resize(n); ntt(a);
-	b.resize(n); ntt(b);
-	vl c(n); ll d = modpow(n, mod-2);
-	rep(i,0,n) c[i] = a[i] * b[i] % mod * d % mod;
-	ntt(c, true); c.resize(s); return c;
+	ntt(inl, rt, rev, n), ntt(inr, rt, rev, n);
+	ll invN = modpow(n, mod - 2);
+	rep(i,0,n) out[-i&(n-1)] = (inl[i] * inr[i] % mod)* invN%mod;
```
simonlindholm: no need for the parentheses. Space before/after %; if that causes the line to exceed 63 chars, you can rename invN to inv

Speaking of renames, we could also rename inl/inr to e.g. L/R, which is harder to typo

content/numerical/NumberTheoreticTransform.h
```diff
@@ -12,43 +12,39 @@
...
+			rep(j,0,k) {
+				ll z = rt[j + k] * a[i + j + k] % mod, &ai = a[i + j];
+				a[i + j + k] = z > ai ? ai - z + mod : ai - z;
+				ai = ai + z >= mod ? ai + z - mod : ai + z;
+			}
 }
-vl conv(vl a, vl b) {
-	int s = sz(a) + sz(b) - 1; if (s <= 0) return {};
-	int L = s > 1 ? 32 - __builtin_clz(s - 1) : 0, n = 1 << L;
-	if (s <= 200) { // (factor 10 optimization for |a|,|b| = 10)
```
simonlindholm: Have you tested whether there exists a cut-off for the new implementation as well? probably isn't necessary any more, but interesting to know.

content/numerical/NumberTheoreticTransform.h
```diff
@@ -12,43 +12,39 @@
...
-	rep(i,1,sz(x)) roots[i] = roots[i-1] * e % mod;
-	ntt(&x[0], &temp[0], &roots[0], sz(x), 1);
+void ntt(vl &a, vl &rt, vl &rev, int n) {
+	rep(i,0,n) if (i < rev[i]) swap(a[i], a[rev[i]]);
+	for (int k = 1; k < n; k *= 2)
+		for (int i = 0; i < n; i += 2 * k)
+			rep(j,0,k) {
+				ll z = rt[j + k] * a[i + j + k] % mod, &ai = a[i + j];
+				a[i + j + k] = z > ai ? ai - z + mod : ai - z;
+				ai = ai + z >= mod ? ai + z - mod : ai + z;
```
simonlindholm: `ai += ai + z >= mod ? z - mod : z;`

content/numerical/NumberTheoreticTransform.h
```diff
@@ -12,43 +12,39 @@
  */
```
simonlindholm: The doc comment still says `May return negative values.`; remove that and mention that inputs have to be in [0, mod) instead.

Update author/date as well.

content/numerical/NumberTheoreticTransform.h
Chillee: Brute force now starts to surpass FFT at N=32.

For N<=4 it's 10x  faster :^)
For N=8 it's 5x faster
For N=16 it's 4x faster
For N=32 it's 1.6x faster
For N=64 it's 1.5x slower.

Definitely not worth it I think.

content/numerical/NumberTheoreticTransform.h
Chillee: I don't like capital variable names :sweat_smile: although your rationale makes sense.

I changed the spacing and renamed `invN` to `inv`. If you think it's worth it to change it to `L/R`, can you do it?

simonlindholm: Not seeing this change. I can make the L/R change; I think it's worth it. (I somewhat like capital letter names, actually - it makes them stand out, and they aren't that much harder to type. I almost always use them for problem inputs.)

Chillee: I also use them for problem inputs, but *only* problem inputs. So it feels weird to me to use them elsewhere. Amended the commit with the actual change...

Chillee: Also made the `L`/`R` change since you're set on it.

Chillee:
uhh.... That was not on purpose.

Chillee:
Ok, I believe I fixed the git disaster - I also fixed a naming issue caused by naming the vector `L` (I renamed the other variable B for "bits", or something like that).

Chillee:
One more note here, I think we might want to add something to make sure `NTT` isn't compiled with `-O3`. Not sure whether we should add something like this: https://stackoverflow.com/questions/31373885/how-to-change-optimization-level-of-one-function or just keep it as documentation.

PR #64, opened 2019-04-22
Shaved another line off SuffixArray.h

Chillee:
Shaved another line off - inspired by looking at my coach's ICPC notebook.

I already ran the fuzz tests (which pass).


simonlindholm:
Nice.

PR #65, opened 2019-04-22
Added coloring to code snippets

Chillee:
Seems like it was just a simple matter of altering `\lstset`.

image

We might want to add a flag or change colors? Not sure how to do that/what colors we should choose.


Tyilo:
I don't know if you have considered it, but we should probably try to match the colors with one the vimrc uses.

Chillee:
Added a flag in `kactlpkg.sty` for controlling colored output and set the default to off.

simonlindholm:
Thanks. Updated to make it changeable from kactl.tex (kactlpkg.sty is for advanced usage), mentioned it in the README, and merged.

PR #67, opened 2019-04-23
Added FFTMod 

Chillee:
I think I've seen a lot of people have this in their library. Allows us to perform FFT over arbitrary moduli without needing to do finicking with primes close to a power of 2 and such. Really, if we ever need more space, I don't think there's anywhere NTT can be used that FFT-MOD can't. An optimized FFT-MOD is only about twice as slow as an optimized NTT. 

On the other hand, NTT is pretty cool, and it is seriously nice to not need to think about precision at all.

Here's an example exemplifying the precision issues (I'll turn it into a fuzz test later). Basically, as long as the coefficients don't surpass ~1e15 FFT is fine. Higher though, and you start to see precision issues.
https://ideone.com/lUrDCT

It is kind of ugly though :'(. I tried my best to shorten it, but it's still uglier than I'd like.

This is somewhat WIP - I still need to add a description about it. But I'd appreciate feedback on the necessity as well as the code itself.


simonlindholm:
> Here's an example exemplifying the precision issues (I'll turn it into a fuzz test later). Basically, as long as the coefficients don't surpass ~1e15 FFT is fine. Higher though, and you start to see precision issues.

So basically, if MOD ~ 1e9, we start seeing precision issues at around n ~ 1e6? That's uncomfortably small :( Especially so since this is at random inputs; the worst case may be, well, worse. I know I researched this before, and most theoretical guarantees were pretty weak -- the FFT currently pessimistically says the following: "For integers rounding works if $(|a| + |b|)\max(a, b) < ~ 10^9$, or in theory maybe $10^6$." And it would take an annoying amount of research to determine whether one can actually get closer to the worst case...

> It is kind of ugly though :'(. I tried my best to shorten it, but it's still uglier than I'd like.

Yeah. :/ Is there any chance of shortening it by making it call the existing conv method?

One alternative, though it loses on code size, is to say that if NTT * 2 ~ FFT-MOD, then can't we run two NTT's and use CRT for the result for about the same performance? It avoids the precision problems, and with luck it might be relatively short since it can reuse a lot of existing KACTL code. (This is currently noted in the FFT description, but the implementation is left as an exercise to the reader.)

> necessity

I don't know, it's tricky making these sorts of decisions... vague preference towards keeping just the NTT and existing FFT description, but it depends on code size and stuff... Though even if we don't include this, it's always nice to have a high-quality implementation that's not included in the pdf, which enterprising teams can uncomment if they think it's important to them.

content/numerical/FastFourierTransform.h
```diff
@@ -49,3 +49,37 @@ vd conv(const vd& a, const vd& b) {
...
+        rep(i, k, 2 * k) rt[i] = rt[i / 2] * z[i & 1];
+    }
+	rep(i,0,a.size()) inl[i] = {a[i] / cut, a[i] % cut};
+	rep(i,0,b.size()) inr[i] = {b[i] / cut, b[i] % cut};
+    fft(inl, rt, rev, n), fft(inr, rt, rev, n);
+	rep(i,0,n){
+        int j = (n - i) & (n - 1);
+        C fl = (inl[i] + conj(inl[j])) * C{0.5, 0}, fs = (inl[i] - conj(inl[j])) * C{0, -0.5},
+          gl = (inr[i] + conj(inr[j])) * C{0.5, 0}, gs = (inr[i] - conj(inr[j])) * C{0, -0.5};
+        outl[-i & (n - i)] = (fl * gl) + (fl * gs) * C{0, 1};
```
simonlindholm: `outl[j] = fl * (gl + gs * C{0, 1});`

Chillee:
I'll check out how a NTT based FFTMod equivalent might look like.

Chillee:
Unluckily, CRT+NTT is 3x slower. Even though our NTT is very fast, the bottleneck isn't the NTT, it's the CRT with 128 bit numbers.

The benchmarks look like this:
```
254ms elapsed [neal]
201ms elapsed [mine]

722ms elapsed [ntt crt]
    283ms elapsed [ntt portion]
    436ms elapsed [crt portion]
```

Chillee:
So, TL;DR: I think we should still include FFT-MOD. I'll think more about whether we can simplify it.

Chillee:
> Yeah. :/ Is there any chance of shortening it by making it call the existing conv method?

Unluckily, I doubt it. Not without significant runtime losses at least (probably 2x). The issue is that if we do a bunch of repeated calls to existing `conv`, we're gonna be redoing a bunch of FFT's (twice for each polynomial).

Chillee:
So I tried it out and it would look look like this
```cpp
template <int MOD> vector<ll> convMod(const vi &a, const vi &b) {
    if (a.empty() || b.empty()) return {};
    ll cut = sqrt(MOD);
    vector<double> ah(sz(a)), al(ah), bh(sz(b)), bl(bh);
    rep(i, 0, sz(a)) ah[i] = a[i] / cut, al[i] = a[i] % cut;
    rep(i, 0, sz(b)) bh[i] = b[i] / cut, bl[i] = b[i] % cut;
    auto h = conv(ah, bh), l = conv(al, bl);
    auto m1 = conv(ah, bl), m2 = conv(al, bh);
    vector<ll> res(h.size());
    for (int i = 0; i < res.size(); i++) {
        ll av = round(h[i]);
        ll bv = round(m1[i]) + round(m2[i]);
        ll cv = round(l[i]);
        av %= MOD, bv %= MOD, cv %= MOD;
        res[i] = (av * cut * cut + bv * cut + cv) % MOD;
    }
    return res;
}
```
Significantly shorter, but still kind of weighty. Also ~2x slower.
```
241ms elapsed [mine]
497ms elapsed [reuse]
```

content/numerical/FastFourierTransform.h
Chillee: Err, these aren't identical. `j=(n-i) & (n-1)`, while we're indexing `-i & (n-i)` at the bottom. Unless those are the same... but I don't think they are.

Chillee:
One thing I did was to move the root computation from the `conv` routines to the `FFT` routine itself. Computing the roots should only take a small percentage (~2.5%) of the total time needed, so this isn't much of a loss. In regular FFT, this should correspond to a 2.5% slowdown since we would be computing roots unnecessarily once. In FFT-MOD this should correspond to a 7.5% slowdown (I actually get 260 vs 236 ms).

I think this is a worthy tradeoff.

simonlindholm:
That seems reasonable. (I imagine it's a larger percentage for small n, though.) I think convMod should probably go in a separate file. Also, can you fix the -Wconversion warnings?

Chillee:
Fixed a bunch of spurious formatting diffs.

content/numerical/FastFourierTransformMod.h
```diff
@@ -0,0 +1,36 @@
...
+	rep(i,0,sz(b)) R[i] = {b[i] / cut, b[i] % cut};
+	fft(L, n, B), fft(R, n, B);
+	rep(i,0,n) {
+		int j = -i & (n - 1);
+		outl[j] = (L[i] + conj(L[j])) * R[i] / (2.0 * n);
+		outs[j] = (L[i] - conj(L[j])) * R[i] / (2.0 * n) / 1i;
+	}
+	fft(outl, n, B), fft(outs, n, B);
+	rep(i,0,sz(res)) {
+		ll av = round(outl[i].real()), cv = round(outs[i].imag());
```
simonlindholm: This has -Wconversion warnings that won't show up unless the template is instantiated. Maybe add a test and make sure that it's warning-free? And/or de-templatize the function.

content/numerical/FastFourierTransformMod.h
```diff
@@ -0,0 +1,36 @@
+/**
+ * Author: chilli
+ * Date: 2019-04-25
+ * License: CC0
+ * Source: http://neerc.ifmo.ru/trains/toulouse/2017/fft2.pdf
+ * Description: Can be used for convolutions modulo arbitrary integers.
+ * Is essentially a higher precision FFT.
+ * Is safe in practice as long as $(|a|+|b|)\sqrt{\max(a)\max(b)} < \mathtt{\sim} 10^{15}$.
```
simonlindholm: `$(|a|+|b|)M < \mathtt{\sim} 10^{15}$`, right? You can shorten the description a bit into "Higher precision FFT, can be used for convolutions modulo arbitrary integers. Safe in practice ..."

simonlindholm: maybe we should change "in practice" to "for random numbers"

content/numerical/FastFourierTransformMod.h
Chillee: Huh, I distinctly remember fixing these issues... but I don't see it in any commits.

content/numerical/FastFourierTransformMod.h
Chillee: I think the claim is a bit stronger than that. I think that statement is a pretty common CP belief.

Proof by consensus, if you will.

Chillee: I think we're good for this issue?

Chillee:
I did some proper benchmarking and it seems like recalculating the roots (with `long double` calculations) slows it down by 25%. If we're only using `double` it slows it down by ~10%.

content/numerical/FastFourierTransform.h
```diff
@@ -3,22 +3,31 @@
...
 typedef complex<double> C;
+typedef complex<long double> Cd;
 typedef vector<double> vd;
-
-void fft(vector<C> &a, vector<C> &rt, vi& rev, int n) {
+void fft(vector<C> &a, vector<C> &rt) {
+	int n = sz(a), L= 32 - __builtin_clz(n); vi rev(n);
+	rep(i,0,n) rev[i] = (rev[i / 2] | (i & 1) << L) / 2;
+	if (rt.empty()) {
+		rt.assign(n, 1);
```
simonlindholm: one way you could shorten this a bit is by taking `k` in the below loop by reference

content/numerical/FastFourierTransform.h
```diff
@@ -3,22 +3,31 @@
...
+		for (int k = 2; k < n; k *= 2) {
+			Cd z[] = {1, polar(1.0, M_PI / k)};
```
simonlindholm: This needs to be `1.0L` to get the long double version of `polar`. Otherwise you end up doing `(double)((ld)a * (ld)b)`, which is if anything *less* precise than `a * b` since it does double rounding. That said, I'm skeptical to whether this can help precision much even if `z` is made to use `complex<long double>`, if `rt` itself isn't.

content/numerical/FastFourierTransform.h
```diff
@@ -27,25 +36,19 @@ void fft(vector<C> &a, vector<C> &rt, vi& rev, int n) {
 			C z(x[0]*y[0] - x[1]*y[1], x[0]*y[1] + x[1]*y[0]);           /// exclude-line
 			a[i + j + k] = a[i + j] - z;
 			a[i + j] += z;
-	}
+		}
 }
-
-vd conv(const vd& a, const vd& b) {
+vd conv(const vd &a, const vd &b) {
```
simonlindholm: KACTL is relatively consistent about using `const vd& a`; let's stick with that (same in convMod)

content/numerical/FastFourierTransform.h
Chillee: I believe my thinking here was to have long double during the multiplication. So, we have a double root, which gets casted to long double during the multiplication and is then converted back to double.

content/numerical/FastFourierTransform.h
Chillee: I think I don't want to complicate the FFT interface much, as it's possible that we might want to use FFT as a subroutine in functions other than `conv`.

simonlindholm: Good point. Another option: make `rt` static, and do
```cpp
static vector<C> rt(2, 1);
for (static int k = 2; k < n; k *= 2) {
    rt.resize(n);
    ... // inner loop as before
}
```
Only wins one line, and can use perf testing (maybe the global variable doesn't get LICM'd), but simplifies the API even further.

content/numerical/FastFourierTransform.h
simonlindholm: Ah, right, complex multiplication is more, ahem, complex than real, so this should indeed win a little. Did you quantify the effects? `polar(1.0L` would still be good. Also, possible -Wconversion warning?

Chillee: So quantifiably, there's reliable precision errors (with random tests) at `1<<20` inputs without the long double, but reliably doesn't have precision errors (ran 20 times) with the long double.

Adding `polar(1.0L` is a little bit annoying, since that would mean I would also need to wrap the `M_PI/k` in a long double too.

simonlindholm: Ah, annoying template stuff... `M_PIl`, if our font can distinguish the `I` from the `l`? :) (Who came up with that name...) Or `1.L * M_PI / k`. Either way I guess we should measure the error compared to directly computed `rt` values; if it doesn't increase precision compared to just ld-precision multiplication there's no point in doing it ofc.

content/numerical/FastFourierTransform.h
Chillee: oh.... smart.

I was gonna ask how it worked with changing `n` sizes, and then I realized.

simonlindholm:
https://github.com/simonlindholm/fft-precision/blob/master/fft-precision.md has a bunch of analysis of FFT precision. I'll try to merge this soon.

Chillee:
:O

PR #69, opened 2019-04-24
Updated Pollard Rho (factor.h)

### LINKED ISSUE #68: Decrease Pollard Rho Length
I suspect that the current Pollard Rho implementation is a lot longer than it needs to be. I've seen some significantly shorter implementations elsewhere.

For example, take this snippet from my coach's notebook: https://github.com/FTRobbin/Dreadnought-Standard-Code-Library/blob/11c861e55a73be9c32a799bfe4398e0c62c2da15/src/%E5%90%AF%E5%8F%91%E5%BC%8F%E5%88%86%E8%A7%A3.cpp

Different API of course, and haven't compared performance. But it may be worth looking at.

Chillee:
I think the current Pollard Rho implementation is A. too long, B. requires Eratosthenes when that seems excessive, and C. can be improved in runtime for large numbers (which is where Pollard Rho matters...). Also, it doesn't have a unit test :^)

This PR fixes all 3 of those issues.
A. 21 lines, down from 35 lines (49 if we count the 14 line sieve dependency).
B. Only requires modmull and MillerRabin
C. Runs about 20% faster for large numbers (~2^50). https://ideone.com/p9KluF It's slower for small numbers (I'm guessing that's because of the sieve), but that's irrelevant for most use cases of PollardRho.

Also added an unit test.

Fixes #68


content/number-theory/factor.h
```diff
@@ -1,52 +1,35 @@
...
-			if (c != d) {
-				res.push_back(c); d /= c;
-				if (d != c) res.push_back(d);
-				break;
-			}
-		}
+ull f(ull x, ull n) { return (mod_mul(x, x, n) + 1) % n; }
+ull Pollard(ull n) {
+	if (isPrime(n)) return n;
+	if (!(n & 1)) return 2;
```
simonlindholm: `auto f = [n](ull x) { return (mod_mul(x, x, n) + 1) % n; };`?

content/number-theory/factor.h
```diff
@@ -1,52 +1,35 @@
...
+	return 0;
 }
+vector<ull> factor(ull n) {
+	if (n==1) return {};
+    ull x = Pollard(n);
+    if (x == n) return {x};
+    auto l = factor(x), r = factor(n / x);
+    l.insert(l.end(), all(r));
+    return l;
+}
```
simonlindholm: newline

content/number-theory/factor.h
```diff
@@ -1,52 +1,35 @@
...
-	pr.assign(all(p));
+	return 0;
 }
+vector<ull> factor(ull n) {
+	if (n==1) return {};
+    ull x = Pollard(n);
+    if (x == n) return {x};
+    auto l = factor(x), r = factor(n / x);
+    l.insert(l.end(), all(r));
+    return l;
```
simonlindholm: mixing tabs and spaces

content/number-theory/factor.h
```diff
@@ -1,52 +1,35 @@
...
-				res.push_back(c); d /= c;
-				if (d != c) res.push_back(d);
-				break;
-			}
-		}
+ull f(ull x, ull n) { return (mod_mul(x, x, n) + 1) % n; }
+ull Pollard(ull n) {
+	if (isPrime(n)) return n;
+	if (!(n & 1)) return 2;
+	rep(i,2,50) {
```
simonlindholm: this can be made an infinite loop, right?

content/number-theory/factor.h
```diff
@@ -1,52 +1,35 @@
...
-}
-void init(int bits) {//how many bits do we use?
-	vi p = eratosthenes_sieve(1 << ((bits + 2) / 3));
-	pr.assign(all(p));
+	return 0;
 }
+vector<ull> factor(ull n) {
+	if (n==1) return {};
+    ull x = Pollard(n);
+    if (x == n) return {x};
```
simonlindholm: `if (isPrime(x)) return {x};` to win a line

content/number-theory/factor.h
```diff
@@ -1,52 +1,35 @@
...
+ * Date: 2019-04-24
  * License: CC0
- * Source: Wikipedia
+ * Source: https://github.com/FTRobbin/Dreadnought-Standard-Code-Library
  * Description: Pollard's rho algorithm. It is a probabilistic factorisation
- * algorithm, whose expected time complexity is good. Before you start using it,
- * run {\tt init(bits)}, where bits is the length of the numbers you use.
- * Returns factors of the input without duplicates.
+ * algorithm, whose expected time complexity is good.
  * Time: Expected running time should be good enough for 50-bit numbers.
```
simonlindholm: `O(n^{1/4})` expected time, IIRC? "good enough for 50-bit numbers" is terribly vague, and probably out of date.

content/number-theory/factor.h
```diff
@@ -1,52 +1,35 @@
...
+ull f(ull x, ull n) { return (mod_mul(x, x, n) + 1) % n; }
+ull Pollard(ull n) {
+	if (isPrime(n)) return n;
+	if (!(n & 1)) return 2;
+	rep(i,2,50) {
+		ull x = i, y = f(x, n), p = __gcd(n + y - x, n);
+		while (p == 1)
+			x = f(x, n), y = f(f(y, n), n), p = __gcd(n + y - x, n);
+		if (p == n) continue;
+		return p;
```
simonlindholm: `if (p != n) return p;`

fuzz-tests/factor.cpp
```diff
@@ -0,0 +1,42 @@
...
+            cout<<i<<endl;
+            assert(isPrime(i));
+        }
+        cur *= i;
+    }
+    assert(cur == N);
+}
+int main() {
+    assert(factor(1).empty());
+    assert(factor(2).size()==1 && factor(2)[0]==2);
```
simonlindholm: I believe `factor(2) == vi{2}` should work, FWIW

fuzz-tests/factor.cpp
```diff
@@ -0,0 +1,42 @@
...
+    rep(n,2,1e5) {
+        auto res = factor(n);
+        isValid(n, res);
+    }
+    rep(i,2,ITERS) {
+        auto res = factor(1 + (uni()%(1ul<<52)));
+        ull cur=1;
```
simonlindholm: Missing `isValid(n, res)`?

content/number-theory/factor.h
Chillee: It can be..., but that would require not using ` rep` loop. 
Like this `    for (int i = 2;; i++) {`

I guess that's better.

content/number-theory/factor.h
Chillee: How?
We need to use the value of `x` after this computation, and we can't replace the first line because passing `n=1` to Pollard doesn't work.

content/number-theory/factor.h
Chillee: It's `O(sqrt(lowest prime factor of composite))`, so `O(N^{1/4})` worst case.

Perhaps it'd be good to state that it's much faster than that on most numbers though.
I'd like to say something like
`On worst case numbers,  `O(N^{1/4})` expected runtime.`
or maybe we can just write the above definition.

content/number-theory/factor.h
```diff
@@ -1,52 +1,35 @@
...
-				x = f(x, d, has);
-				y = f(f(y, d, has), d, has);
-			}
-			if (c != d) {
-				res.push_back(c); d /= c;
-				if (d != c) res.push_back(d);
-				break;
-			}
-		}
+ull f(ull x, ull n) { return (mod_mul(x, x, n) + 1) % n; }
```
simonlindholm: remove this

content/number-theory/factor.h
```diff
@@ -1,52 +1,35 @@
...
+	if (isPrime(n)) return n;
+	if (!(n & 1)) return 2;
+	for (ull i = 2;; i++) {
+		ull x = i, y = f(x), p = __gcd(n + y - x, n);
+		while (p == 1)
+			x = f(x), y = f(f(y)), p = __gcd(n + y - x, n);
+		if (p != n) return p;
 	}
-	return res;
+	return 0;
```
simonlindholm: no need for this given the inf loop

content/number-theory/factor.h
```diff
@@ -1,52 +1,35 @@
...
-			}
-			if (c != d) {
-				res.push_back(c); d /= c;
-				if (d != c) res.push_back(d);
-				break;
-			}
-		}
+ull f(ull x, ull n) { return (mod_mul(x, x, n) + 1) % n; }
+ull Pollard(ull n) {
+	auto f = [&n](ull x) { return (mod_mul(x, x, n) + 1) % n; };
```
simonlindholm: don't capture n by reference unnecessarily

content/number-theory/factor.h
simonlindholm: ah, err, I meant `if (isPrime(n)) return {n};` (just before the call to Pollard, so that function doesn't have to check isPrime)

Chillee: Ah I see.

content/number-theory/factor.h
Chillee: Ah interesting. Didn't know that gcc wouldn't throw an error if it could tell it was an infinite loop.

Chillee:
Oh, one more note.  This changes the API a little bit compared to the old implementation. The old one would return a vector of all unique prime factors, while this one will return a prime factor multiple times.

ie: old one would give `{5}` for 125, while the new one would return `{5,5,5}`. I'd err on the side of returning multiple copies if they exist. It's trivial to go from `{5,5,5}` -> `{5}` while modifying the old pollard-rho to return all prime factors would be more difficult. Since they're useful for different tasks, I think this is the preferable API.

simonlindholm:
Agreed, the new behavior seems much less surprising. I made a note in the description.

Issue #70, opened 2019-04-24
Improve ModMull


Chillee:
I don't really understand why we have `mod_mul`. If we're not supporting 32 bit (ie: CF), can't we juts use `__int128_t`?

Alternately, we can use the equivalent of this: https://github.com/RamchandraApte/OmniTemplate/blob/master/modulo.h#L23

The current KACTL one is slower (by 5-20x), has that annoying `bit` thing, and is longer.

simonlindholm:
Let's close this now, I guess?

PR #71, opened 2019-04-24
Updated ModMul

Chillee:
See #70.
This new one has the following advantages
1. Doesn't require finicking with some `bits` parameter.
2. Is shorter
3. Is faster by more than an order of magnitude: https://ideone.com/72x8RQ
4. Has ~2x more throughput than `__int128_t` multiplication, while having ~`40%` less latency


content/number-theory/ModMulLL.h
```diff
@@ -1,25 +1,19 @@
 /**
- * Author: Lukas Polacek
- * Date: 2010-01-26
- * License: CC0
- * Source: TopCoder tutorial
+ * Author: chilli, Ramchandra Apte
+ * Date: 2019-04-24
+ * License: MIT
+ * Source: https://github.com/RamchandraApte/OmniTemplate/blob/master/modulo.h
+ * Works for numbers < 2^63.
```
simonlindholm: Move this to the description (replacing the vague "large $c$")

simonlindholm:
> Has ~2x more throughput than __int128_t multiplication, while having 40% more latency :(

40% less, actually: https://ideone.com/idxuvN
Also, using double instead of long double (which is fine for <= 53-bit numbers) decreases that by another 20% and throughput by 30%: https://ideone.com/yDYAet

(All numbers approximate; microbenchmarking is hard.)

@RamchandraApte do you know why this (according to all tests I've done) works for moduli up to 2^63? I can prove it for up to 2^62, but for values closer to 2^63 I don't understand why the `(ret >= (ll)M)` check potentially breaking doesn't affect anything. Also, any chance we can get this as CC0-licensed rather than MIT? :) Requiring attribution is annoying in competitive programming, because it's unclear whether you need to do it during contests.

RamchandraApte:
Multiplication and division adds at most 2^-(63+1) relative error (63 is the number of mantissa bits in 80-bit floating point) since it produces the nearest floating point number. Therefore the relative error of a*b/M is at most 2^-63. Since |a*b/M| < M <= 2^63, therefore the absolute error is less than 1. Therefore the the absolute error in the truncated value is at most 1.

I'm not an expert on floating point so I could be wrong though.

simonlindholm:
Yep, that's my analysis as well (with a bit of extra math to account for two relative errors of 2^-64 adding up to slightly more than 2^-63 relative error). What I don't understand is why overflow isn't an issue in `(ret >= (ll)M)`.

RamchandraApte:
I hereby license the following code (taken from [here](https://github.com/RamchandraApte/OmniTemplate/blob/164d5a83527bacdab4d5007cdcdfdb176c5af1c1/modulo.h#L22)) under the terms of Creative Commons CC0 v1.0, in addition to the MIT license.

```
md operator*(md co& a, md co& b){
	const ull quot = ld(a.x)*ld(b.x)/ld(M);
	// Computes the approximate remainder
	const ll rem = ull(a.x)*ull(b.x)-ull(M)*quot;
	if(rem < 0){return {rem+M, no_mod{}};}
	if(rem >= M){return {rem-M, no_mod{}};}
	return {rem, no_mod{}};
}
```

Let me know if you need anything more but I think this is good enough.

simonlindholm:
That's perfect, thanks! :)

Chillee:
If we're gonna update `mod_pow`, we might as well convert it to iterative. Runs about `30%` faster and is less characters.

simonlindholm:
> What I don't understand is why overflow isn't an issue in `(ret >= (ll)M)`.

Okay, I got a proof for this. It's trickier than expected.

Let `round(x)` denote rounding `x` to the nearest long double, `frac(x) = x - floor(x)`, and assume `0 <= a, b < c <= 2^63`. What we want to prove is that `ab - floor(round(round(ab) / c))*c < 2^63` always holds. Let's assume the opposite for sake of contradiction.

If `round(ab) / c >= ab / c` we're fine: `floor(round(round(ab) / c)) >= floor(round(ab / c)) >= floor(ab / c)`, since floor and round is monotonic, and since `ab / c < c^2/c <= 2^63`, the integer `floor(ab / c)` is representable, so `round` can't skip past it. Thus, `ab - floor(round(round(ab) / c))*c <= ab - floor(ab / c)*c = ab % c < c <= 2^63`, contradiction.

Otherwise, `round(ab) / c < ab / c < c^2/c <= 2^63`. Let k be such that `2^k <= round(ab) / c < 2^(k+1)`; then `k <= 62`.

If `round(round(ab) / c)` rounds upward to an integer, then `ab - floor(round(round(ab) / c))*c = ab - round(round(ab) / c)*c <= ab - round(ab) / c * c = ab - round(ab) <= ab * 2^-64 < c^2/2^64 < 2^63`, contradiction.

Otherwise, since `round(round(ab) / c)` only has 64 bits of precision and `round` rounds to nearest, breaking ties towards even, `frac(round(ab) / c) < 1 - 2^(k-64)`. (This is the magic part.)

Division can't move us below an integer, so `floor(round(round(ab) / c)) = floor(round(ab) / c)`. Thus we can rewrite our inequality as

`ab - floor(round(ab) / c)*c >= 6^63`
`ab - ((round(ab) / c) - frac(round(ab) / c))*c >= 6^63`
`ab - round(ab) >= 6^63 - frac(round(ab) / c)*c`

Since `round(ab) < c * 2^(k+1) <= 2^k * 2^64`, `round(ab)` rounds to an ulp of at most `2^k`, and as such `ab - round(ab) <= 2^(k-1)`. Combining with the above, we derive

`2^(k-1) >= ab - round(ab) >= 6^63 - frac(round(ab) / c)*c > 2^63 - (1 - 2^(k-64))*c`
which rearranges to
`2c + (1 - c/2^63)*2^k > 2^64`

Noting that `1 - c/2^63` is always non-negative, and using `k < 63`, this implies
`2c + (1 - c/2^63)*2^63 > 2^64`
or `c > 2^63`, which is a contradiction.

simonlindholm:
Oh, and the proof that `ab - floor(round(round(ab) / c))*c ∈ [-c, 2c)`: we have `|round(x) - x| ≤ x * 2^-64`, we can write each `round(x)` as `x * (1 + ε/2^-64)` where `|ε| ≤ 1`. Also, let's write `floor(x)` as `x - ε` with `0 ≤ ε < 1`.

Then we can rewrite:
`ab - floor(round(round(ab) / c))*c ∈ [-c, 2c)`
`floor(round(round(ab) / c)) ∈ ab/c + (-2, 1]`
`round(round(ab) / c) - ε ∈ ab/c + (-2, 1]`
⇐ `round(round(ab) / c) ∈ ab/c + [-1, 1]`
`ab / c (1+ε₁/2^-64)(1+ε₂/2^-64) ∈ ab/c + [-1, 1]`
`ab / c (ε₁/2^-64 + ε₂/2^-64 + ε₁ε₂/2^-128) ∈ [-1, 1]`
`ab / c |ε₁/2^-64 + ε₂/2^-64 + ε₁ε₂/2^-128| ≤ 1`
⇐ `ab / c (2/2^-64 + 1/2^-128) ≤ 1`
`ab / c ≤ 1 / (1/2^-63 + 1/2^-128)`
⇐ `(c-1)(c-1)/c ≤ 1 / (1/2^-63 + 1/2^-128)`
`c - 2 + 1/c ≤ 1 / (1/2^-63 + 1/2^-128)`
`c - 1 ≤ 1 / (1/2^-63 + 1/2^-128)`
`c - 1 ≤ 2^63 * (1 - 2^-65 / (1 + 2^-65))`
⇐ `c - 1 ≤ 2^63 * (1 - 2^-65)`
`c - 1 ≤ 2^63 - 2^-2`
which holds for `c ≤ 2^63`.

simonlindholm:
TBH I'm not actually sure we need this at all, when it's relatively rare, and int128 multiplication also works and is roughly as fast. Still, it's an improvement from before, so merged.

Chillee:
I mean, this function is substantially faster than `__int128_t`, and is also super short. Plus it has the auxiliary benefit of working on 32 bit machines too.

PR #73, opened 2019-04-25
Added hashing

Chillee:
Done in a similar fashion to MIT's. Unlike MIT's, I elected to not do section hashing, and to simply hash everything that's printed.

Taking ecnerwala's suggestion, I believe I moved the hashing to the last pass of the preprocessor. I also decided to put first 5 chars of the hash in the header, which looks good imo.
image

Sometimes you can run into issues like this though: 
image

This might just be a quirk of my latex compilation, as it seems like it's possible for this to also overlap with the line number count as well.

### DIFF
--- a/build/kactlprocessor.py
+++ b/build/kactlprocessor.py
@@ -7,6 +7,7 @@
 from __future__ import print_function
 import sys
 import getopt
+import subprocess
 
 
 def escape(input):
@@ -128,6 +129,10 @@ def processwithcomments(caption, instream, outstream, listingslang = None):
         nsource = nsource.rstrip() + source[end:]
     nsource = nsource.strip()
 
+    hash_script = 'hash' if listingslang else 'hash-cpp'
+    p = subprocess.Popen(['sh', '../content/contest/%s.sh' % hash_script], stdin=subprocess.PIPE, stdout=subprocess.PIPE)
+    hsh, _ = p.communicate(nsource)
+    hsh = hsh.split(None, 1)[0][:5]
     # Produce output
     out = []
     if warning:
@@ -147,7 +152,7 @@ def processwithcomments(caption, instream, outstream, listingslang = None):
         if includelist:
             out.append(r"\leftcaption{%s}" % pathescape(", ".join(includelist)))
         if nsource:
-            out.append(r"\rightcaption{%d lines}" % len(nsource.split("\n")))
+            out.append(r"\rightcaption{%s, %d lines}" % (hsh, len(nsource.split("\n"))))
         langstr = ", language="+listingslang if listingslang else ""
         out.append(r"\begin{lstlisting}[caption={%s}%s]" % (pathescape(caption), langstr))
         out.append(nsource)
--- a/build/kactlprocessor.py
+++ b/build/kactlprocessor.py
@@ -129,10 +129,10 @@ def processwithcomments(caption, instream, outstream, listingslang = None):
         nsource = nsource.rstrip() + source[end:]
     nsource = nsource.strip()
 
-    hash_script = 'hash' if listingslang else 'hash-cpp'
+    hash_script = 'hash'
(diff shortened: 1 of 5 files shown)


content/contest/hash.sh
```diff
@@ -0,0 +1 @@
+tr -d '[:space:]' | md5sum
```
simonlindholm: I think we can get rid of this; hash-cpp.sh should be fine for other file types as well in 99% of cases, and when it isn't (e.g. Python int division accidentally commenting out a line; not that KACTL currently supports Python) we can skip the hash for that file.

content/contest/.vimrc
```diff
@@ -1,2 +1,5 @@
 set cin aw ai is ts=4 sw=4 tm=50 nu noeb bg=dark ru cul
 sy on   |   im jk <esc>   |   im kj <esc>   |   no ; :
+
+com -range=% -nargs=1 P exe "<line1>,<line2>!".<q-args> |y|sil u|echom @"
+au FileType cpp com! -buffer -range=% Hash <line1>,<line2>P cpp -P -fpreprocessed | tr -d '[:space:]' | md5sum | cut -c1-5
```
simonlindholm: remove `au FileType cpp`, and inline the `:P` command since it's not generally that useful

content/contest/hash-cpp.sh
```diff
@@ -0,0 +1 @@
+cpp -P -fpreprocessed | tr -d '[:space:]' | md5sum
```
simonlindholm: newline at end of file, and add a `cut`?

simonlindholm: add this to chapter.tex, with a description of how to use it and that it ignores comments and whitespace

build/kactlprocessor.py
```diff
@@ -128,6 +129,10 @@ def processwithcomments(caption, instream, outstream, listingslang = None):
         nsource = nsource.rstrip() + source[end:]
     nsource = nsource.strip()
 
+    hash_script = 'hash' if listingslang else 'hash-cpp'
```
simonlindholm: maybe skip hashing for .txt, .sh and .py?

simonlindholm:
> This might just be a quirk of my latex compilation, as it seems like it's possible for this to also overlap with the line number count as well.

Yep. Mostly it don't happen because the Time description is short; it should be easy to reword the affected descriptions so that it doesn't happen.

build/kactlprocessor.py
Chillee: We already don't do hashing for those, since (at least for .sh and .txt), they're not processed by `processwithcomments`, they're processed by `processraw`.

Regardless, in according with the other comment, I'll probably just hash everything with `hash-cpp`.

Chillee:
> also, maybe rename the command hash?
User defined commands have to start with a capital letter.

Chillee:
Hmm, maybe there should just be a `README` to this document - there's also stuff like the vim keybindings that seem good to document (but are kind of awkward to do so in the document itself).

simonlindholm:
> User defined commands have to start with a capital letter.

Hm, that's annoying... maybe we can use keybindings instead?
```
cmap H w !cpp -P -fpreprocessed \| tr -d '[:space:]' \| md5sum \| cut -c1-6
```
Loses the property that `:Hash` hashes the whole document, but I can live without that.

> Hmm, maybe there should just be a README to this document - there's also stuff like the vim keybindings that seem good to document (but are kind of awkward to do so in the document itself).

Maybe... I don't know... could also put text in chapter.tex, if that helps...

Chillee:
> Hm, that's annoying... maybe we can use keybindings instead?
That works nicely. It still works for hashing the entire file too.

> Maybe... I don't know... could also put text in chapter.tex, if that helps...
I just put a note in the `.vimrc` that it's about hashing.

I think at some point it'd be a good idea to write a README  for teams looking to use this document. Stuff like justifying macros, explaining certain decisions, specific flags (for example, colored code or not), as well as providing a list of algorithms we haven't included in the main document. For now though, I think what we have is good enough.

simonlindholm:
Agreed, this looks good. Merged.

Issue #23, opened 2017-05-31
Usability improvements for numeric integration


simonlindholm:
- ~~should take a functor, not a function pointer~~ (done in #74)
- should only return once `b - a` is smaller than 1/k times the original interval size

PR #74, opened 2019-04-25
Switched integrate functions to use functors

Chillee:
See https://github.com/kth-competitive-programming/kactl/issues/23

I'm not really sure what the other point in that issue is talking about (what's k?). Plus, it seems like a pain to do :thinking: 

I checked the example and such to make sure it still compiles and works (shouldn't have changed any functionality...)

### DIFF
--- a/content/numerical/IntegrateAdaptive.h
+++ b/content/numerical/IntegrateAdaptive.h
@@ -19,8 +19,7 @@ typedef double d;
 template <class F>
 d rec(F f, d a, d b, d eps, d S) {
 	d c = (a + b) / 2;
-	d S1 = S(a, c);
-	d S2 = S(c, b), T = S1 + S2;
+	d S1 = S(a, c), S2 = S(c, b), T = S1 + S2;
 	if (abs(T - S) <= 15 * eps || b - a < 1e-10)
 		return T + (T - S) / 15;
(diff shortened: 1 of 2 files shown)


content/numerical/IntegrateAdaptive.h
```diff
@@ -15,18 +15,21 @@
 #pragma once
 
 typedef double d;
-d simpson(d (*f)(d), d a, d b) {
+template<class F>
+d simpson(F f, d a, d b) {
 	d c = (a+b) / 2;
 	return (f(a) + 4*f(c) + f(b)) * (b-a) / 6;
 }
```
simonlindholm: We can shorten this a bit (and in particular get rid of the `template<class T>` bit) by making the function into a macro:
```
#define S(a,b) (f(a) + 4*f((a+b) / 2) + f(b)) * (b-a) / 6
```

simonlindholm:
> I'm not really sure what the other point in that issue is talking about (what's k?). Plus, it seems like a pain to do :thinking:

k is a user-provided parameter -- the idea is that integrateAdaptive might be fooled into thinking that the function e.g. follows a straight line because it only samples 4 points, while if we split the interval into larger parts then it would notice the non-straight-line parts of the function. It should only add O(k) additional work. We can leave that for another PR, of course. Could be done as a loop by the caller as well, but it's a sharp edge that's easy to miss.

PR #76, opened 2019-04-25
Made Manachers return the vector that it's calculating

Chillee:
Currently Manacher's doesn't return the vector it calculates.


simonlindholm:
Lol

Issue #77, opened 2019-04-26
Porting CP Geo Handbook to KACTL/Other Geo additions


Chillee:
https://codeforces.com/blog/entry/56773
https://vlecomte.github.io/cp-geo.pdf

My main goal in doing this would be the following:
1. Add more tests for the geometry functions.
2. Reduce verbosity. Vlecomte's book has a lot of nice implementations, so I'll be picking the better one between the two.
3. Augment with some new functions. Namely, Halfplane.

For the sake of consistency (and because mostly there's not much of a difference...) I will be using the current KACTL Point class for everything.

- [x] Point.h
- [x] lineDistance.h (we could change it to `lineSqDist` so that it was exact for integers)
- [x] SegmentDistance.h 
- [x] SegmentIntersection.h
- [x] SegmentIntersectionQ.h
- [x] lineIntersection.h
- [x] sideOf.h
- [x] onSegment.h
- [x] linearTransformation.h
- [x] Angle.h
- [x] CircleIntersection.h Maybe update API to allow for returning 1.
- [ ] circleTangents.h
- [x] circumcircle.h
- [x] ~MinimumEnclosingCircle.h~
- [x] insidePolygon.h
- [x] PolygonArea.h
- [x] ~PolygonCenter.h~
- [x] ~PolygonCut.h~
- [x] ~ConvexHull.h~
- [x] ~PolygonDiameter.h~
- [x] ~PointInsideHull.h~
- [x] ~LineHullIntersection.h~
- [x] ~ClosestPair.h~
- [x] ~kdTree.h~
- [x] ~FastDelaunay.h~
- [x] PolyhedronVolume.h
- [x] Point3D.h
- [x] ~3dHull.h~
- [x] sphericalDistance.h

Potential new implementations that KACTL doesn't have (may move to separate issue):
- [ ] Circle-Line intersection
- [ ] Convex-Convex polygon intersection in `log n` time
- [ ] Max distance across convex polygon in `log n` time

Chillee:
I have resolved all the issues/have a PR open, or plan on making a new issue for the new additions/major API changes (circleTangents.h).

PR #78, opened 2019-04-26
Updated onSegment.h and both segment intersections

Chillee:
See #77.

This is a good win. Naively, we only use 24 lines now, compared to 50 lines previously. However, that isn't completely fair, as this new one introduces more dependencies. Even including all dependencies however, the functions required for the new `SegmentIntersection.h` total 21 lines, while the previous function was 27 lines.

I also shortened the function names (`segmentIntersection` is too verbose), in line with how most modern KACTL code is written.

I also changed the API a bit for SegmentIntersection.h. Instead of returning `cnt` and up to 2 points in separate variables, the new API simply returns a vector with 0,1, or 2 elements.

I deleted `SegmentIntersectionQ.h`, as I don't think there's much point in it anymore. Previously, `SegmentIntersection.h` was extremely verbose and had a somewhat cumbersome API where you'd need to create 2 dummy points even if you didn't want them. For the new API, you can use it very similarly to the old `SegmentIntersectionQ.h` if desired.

`sz(segInter(a,b,c,d))>0` is equivalent to the old `segmentIntersectionQ(a,b,c,d)`.

I also added a fuzz test trying a bunch of integer points on a 5x5 grid. I think trying integer points is the right thing to do, as an important part of this function is handling the "degenerate cases" correctly.

I also submitted to https://open.kattis.com/problems/segmentintersection and passed it with the new code: https://ideone.com/xT9jXo

Primary points of contention:
1. The new API change. I think it's better - is there any reason you'd have for preferring the old one?
2. I also added a function to `Point.h`: `sgn(x)`. Simply returns the `sign` of a value, with some epsilon handling. My coach strongly suggested this as a good way of dealing with epsilons. I'm not sure what the philosophy is in KACTL for how to deal with epsilon issues in geometry, but I think this is a solid method.


simonlindholm:
> The new API change. I think it's better - is there any reason you'd have for preferring the old one?

The new version is definitely better. Easy to change to return a bool as well.

simonlindholm:
Argh, pressed enter too early.

> I also added a function to Point.h: sgn(x). Simply returns the sign of a value, with some epsilon handling. My coach strongly suggested this as a good way of dealing with epsilons. I'm not sure what the philosophy is in KACTL for how to deal with epsilon issues in geometry, but I think this is a solid method.

This is my main worry as well. Does one really want epsilon checks everywhere? I'm not sure... I'll need to ponder this.

Additionally, using `sgn` everywhere might be slightly slower than direct comparisons. (And it introduces a -Wconversion warning with `ll`, but then, it's easy to change to `(x > 0) - (x < 0)` when typing it in by hand.)

Chillee:
If we want to handle the `-Wconversion` issue, it's easy enough to just do
`(x>T(-EPS)) - (x<T(EPS))` as well.

content/geometry/Point.h
```diff
@@ -9,6 +9,7 @@
  */
 #pragma once
 
+template <typename T> int sgn(T x) { return (x > 0) - (x < 0);}
```
simonlindholm: `template<class T>` for consistency (and it's shorter so you can add back the space at the end)

content/geometry/SegmentIntersection.h
```diff
@@ -1,53 +1,46 @@
...
+ * Date: 2019-04-27
  * License: CC0
- * Source:
+ * Source: https://vlecomte.github.io/cp-geo.pdf
  * Description:\\
 \begin{minipage}{75mm}
-If a unique intersetion point between the line segments going from s1 to e1 and from s2 to e2 exists r1 is set to this point and 1 is returned.
-If no intersection point exists 0 is returned and if infinitely many exists 2 is returned and r1 and r2 are set to the two ends of the common line.
+If a unique intersection point between the line segments going from s1 to e1 and from s2 to e2 exists then it is returned.
+If no intersection point exists an empty vector is returned. If infinitely many exist a vector with 2 elements is returned.
```
simonlindholm: ", containing the endpoints of the common line segment"?

content/geometry/SegmentIntersection.h
```diff
@@ -1,53 +1,46 @@
...
 \includegraphics[width=\textwidth]{content/geometry/SegmentIntersection}
 \end{minipage}
- * Status: Well tested with unitTest and with Kattis problem intersection.
+ * Status: Well tested with fuzz-test and with Kattis problem intersection.
  * Usage:
- * Point<double> intersection, dummy;
- * if (segmentIntersection(s1,e1,s2,e2,intersection,dummy)==1)
- *   cout << "segments intersect at " << intersection << endl;
+ * vector<P> inter = segInter(s1,e1,s2,e2);
+ * if (inter.size()==1)
```
simonlindholm: `sz(inter)`

content/geometry/SegmentIntersection.h
```diff
@@ -1,53 +1,46 @@
...
-template<class P>
-int segmentIntersection(const P& s1, const P& e1,
-		const P& s2, const P& e2, P& r1, P& r2) {
-	if (e1==s1) {
-		if (e2==s2) {
-			if (e1==e2) { r1 = e1; return 1; } //all equal
-			else return 0; //different point segments
-		} else return segmentIntersection(s2,e2,s1,e1,r1,r2);//swap
+template <class P>
+bool segInterProper(P a, P b, P c, P d, P& out) {
```
simonlindholm: this function adds a lot of boilerplate, can you inline it into the caller? Also, let's add a comment about when this code path returns true, to make this code simpler to modify (I imagine you'll want to make it return just true/false most of the time, and then it's nice to know when you can leave out the onSegment checks).

content/geometry/SegmentIntersection.h
```diff
@@ -1,53 +1,46 @@
...
+}
+template<class P> vector<P> segInter(P a, P b, P c, P d) {
+	P out;
+	if (segInterProper(a, b, c, d, out)) return {out};
+	set<P> s;
+	if (onSegment(c, d, a)) s.insert(a);
+	if (onSegment(c, d, b)) s.insert(b);
+	if (onSegment(a, b, c)) s.insert(c);
+	if (onSegment(a, b, d)) s.insert(d);
+	return vector<P>(all(s));
```
simonlindholm: `return {all(s)}` should work I think

content/geometry/onSegment.h
```diff
@@ -1,17 +1,15 @@
...
 #pragma once
 
 #include "Point.h"
 
-template<class P>
-bool onSegment(const P& s, const P& e, const P& p) {
-	P ds = p-s, de = p-e;
-	return ds.cross(de) == 0 && ds.dot(de) <= 0;
+template <class P> bool onSegment(P s, P e, P p) {
+	return s.cross(e, p) == 0 && (s - p).dot(e - p) <= 0;
```
simonlindholm: make this `p.cross(s, e) == 0` to improve performance slightly

content/geometry/SegmentIntersection.h
```diff
@@ -1,53 +1,46 @@
...
-		if (e2==s2) {
-			if (e1==e2) { r1 = e1; return 1; } //all equal
-			else return 0; //different point segments
-		} else return segmentIntersection(s2,e2,s1,e1,r1,r2);//swap
+template <class P>
+bool segInterProper(P a, P b, P c, P d, P& out) {
+	double oa = c.cross(d, a), ob = c.cross(d, b),
+	       oc = a.cross(b, c), od = a.cross(b, d);
+	if (sgn(oa) * sgn(ob) < 0 && sgn(oc) * sgn(od) < 0) {
+		out = (a * ob - b * oa) / (ob - oa);
```
simonlindholm: (very nice to see the returned value become just a division without any further translations -- then the result is precise when inputs are integers, so you can put intersections in sets etc.)

content/geometry/SegmentIntersection.h
```diff
@@ -1,53 +1,46 @@
...
  * License: CC0
- * Source:
+ * Source: https://vlecomte.github.io/cp-geo.pdf
  * Description:\\
 \begin{minipage}{75mm}
-If a unique intersetion point between the line segments going from s1 to e1 and from s2 to e2 exists r1 is set to this point and 1 is returned.
-If no intersection point exists 0 is returned and if infinitely many exists 2 is returned and r1 and r2 are set to the two ends of the common line.
+If a unique intersection point between the line segments going from s1 to e1 and from s2 to e2 exists then it is returned.
+If no intersection point exists an empty vector is returned. If infinitely many exist a vector with 2 elements is returned.
 The wrong position will be returned if P is Point<int> and the intersection point does not have integer coordinates.
```
simonlindholm: `Point<int>` ->` Point<ll>` while you're changing things, no-one should have reason to use `Point<int>` anymore

simonlindholm:
Also, I'm happy you changed `sgn` back to not use epsilons. The current behavior is sane for floats (if my thinking is correct it returns a result consistent with intersection of the input points moved around by epsilon); making it prefer special cases ("colinear is more likely than almost colinear") would require epsilons in a bunch more places than just the sgn's, and the epsilons would depend on the input coordinates and vary between sgn usages.

content/geometry/onSegment.h
Chillee: Why would that improve performance?

content/geometry/SegmentIntersection.h
Chillee: I might have a preference to keep this as its own function - I think it's nice to use it on its own sometimes.

On the other hand, I guess it isn't that hard to do 
```
auto res = segInter(a,b,c,d);
if (sz(res) == 1) do something with res[0]
```

Chillee: I changed it.

content/geometry/onSegment.h
simonlindholm: Because then `s - p` and `s - e` can be used both here and in the dot call. Probably a minuscule difference but still.

content/geometry/SegmentIntersection.h
simonlindholm: I suspect it's not very useful on its own, given that it doesn't include line segment endpoints. And it's easy enough to modify/use segInter.

content/geometry/SegmentIntersection.h
```diff
@@ -16,31 +16,24 @@ Products of three coordinates are used in intermediate steps so watch out for ov
...
-		return true;
-	}
-	return false;
-}
 template<class P> vector<P> segInter(P a, P b, P c, P d) {
-	P out;
-	if (segInterProper(a, b, c, d, out)) return {out};
+	auto oa = c.cross(d, a), ob = c.cross(d, b),
+	     oc = a.cross(b, c), od = a.cross(b, d);
+	// Returns if only one intersection exists
```
simonlindholm: This comment is not quite true; the one-intersection case can also occur if one of the endpoints lines on the other line. Rather, this finds intersections of open, non-parallel line segments.

PR #81, opened 2019-04-26
Updated `inPolygon`

Chillee:
Built on top of #78 , so review/merge that one first. I'll rebase when that's merged.

This is currently mostly just for keeping track of what I'm working on.


content/geometry/insidePolygon.h
```diff
@@ -1,34 +1,29 @@
...
-		n += (max(i->y,j->y) > p.y && min(i->y,j->y) <= p.y &&
-				((*j-*i).cross(p-*i) > 0) == (i->y <= p.y));
-	}
-	return n&1; //inside if odd number of intersections
+template<class P>
+bool inPolygon(vector<P> &p, P a, bool strict = true) {
+		int nCross = 0, n = sz(p);
+		rep(i, 0, n) {
+			if (onSegment(p[i], p[(i + 1) % n], a)) return !strict;
+			nCross += (sgn(a.y >= p[(i + 1) % n].y) -
```
simonlindholm: sgn(bool) looks off, was this meant as sgn(everything) > 0?

Chillee: You're right that there should probably be a `sgn` around `a.cross`. Take a look at page 61 here for an explanation of the method: https://vlecomte.github.io/cp-geo.pdf

Chillee: I think I updated it - should resolve issues.

Chillee:
The logic for checking whether it crosses the line is fundamentally the same, so I don't think the behavior for numerical accuracy should be better/worse. It's mainly just a shortened version (with the observation that subtracting the booleans should result in the same sign as the cross product if it intersects).

image

Chillee:
Kattis PointInPolygon submission: https://ideone.com/T4D6qk

Couldn't find `copsrobbers` or `intersection`.

content/geometry/onSegment.h
```diff
@@ -4,6 +4,7 @@
  * License: CC0
  * Source: https://vlecomte.github.io/cp-geo.pdf
  * Description: Returns true iff p lies on the line segment from s to e. Intended for use with e.g. Point<ll> where overflow is an issue.  Use \texttt{(segDist(s,e,p)<=epsilon)} instead when using Point<double>.
+ * Works with integers.
```
simonlindholm: There's already a comment about about integers/doubles; I don't think this adds anything. (That comment could be reformulated to drop the "e.g." and "where overflow is an issue"; `Point<int>` should die)

content/geometry/insidePolygon.h
```diff
@@ -1,34 +1,31 @@
...
-	}
-	return n&1; //inside if odd number of intersections
+template<class P>
+bool inPolygon(vector<P> &p, P a, bool strict = true) {
+		int cnt = 0, n = sz(p);
+		rep(i,0,n) {
+			P np = p[(i + 1) % n];
+			if (onSegment(p[i], np, a)) return !strict;
+			//or: if (segDist(p[i], np, a) <= eps) return !strict;
+			cnt += ((a.y>=np.y)-(a.y>=p[i].y))*a.cross(p[i],np) > 0;
```
simonlindholm: `cnt ^=`, to get rid of the bitand at the end

content/geometry/insidePolygon.h
```diff
@@ -1,34 +1,31 @@
...
-		//increment n if segment intersects line from p
-		n += (max(i->y,j->y) > p.y && min(i->y,j->y) <= p.y &&
-				((*j-*i).cross(p-*i) > 0) == (i->y <= p.y));
-	}
-	return n&1; //inside if odd number of intersections
+template<class P>
+bool inPolygon(vector<P> &p, P a, bool strict = true) {
+		int cnt = 0, n = sz(p);
+		rep(i,0,n) {
+			P np = p[(i + 1) % n];
```
simonlindholm: we could do `np = p[(i ?: n) - 1]` to get rid of the modulo

content/geometry/insidePolygon.h
simonlindholm: also, I'd appreciate some more whitespace here (e.g. around the `*`), I think it's hard to follow where parentheses end. I think you could win some characters by converting `>=` to `<`? and by renaming `np` to `q`, if needed

content/geometry/insidePolygon.h
```diff
@@ -1,34 +1,31 @@
 /**
- * Author: Ulf Lundstrom
- * Date: 2009-03-22
+ * Author: Victor Lecomte, chilli
+ * Date: 2019-04-26
  * License: CC0
- * Source: Basic geometry
- * Description: Returns true if p lies within the polygon described by the points between iterators begin and end. If strict false is returned when p is on the edge of the polygon. Answer is calculated by counting the number of intersections between the polygon and a line going from p to infinity in the positive x-direction. The algorithm uses products in intermediate steps so watch out for overflow. If points within epsilon from an edge should be considered as on the edge replace the line "if (onSegment..." with the comment bellow it (this will cause overflow for int and long long).
+ * Source: https://vlecomte.github.io/cp-geo.pdf
+ * Description: Returns true if p lies within the polygon. If p is on the edge of the polygon, it returns false when strict is true. Answer is calculated by counting the number of intersections between the polygon and a line going from p to infinity in the positive x-direction. The algorithm uses products in intermediate steps so watch out for overflow. Works with integers.
```
simonlindholm: again, the `Works with integers.` comment is pretty much implicit. The interesting question is whether it works with doubles. It seems to have some special-casing for it, but I'm not quite sure about the handling of parallel lines, I'd need to think about it.

content/geometry/insidePolygon.h
```diff
@@ -1,34 +1,31 @@
...
+ * Status: fuzz-tested and Kattis problem pointinpolygon
  * Usage:
- * 	typedef Point<int> pi;
- * 	vector<pi> v; v.push_back(pi(4,4));
- * 	v.push_back(pi(1,2)); v.push_back(pi(2,1));
- * 	bool in = insidePolygon(v.begin(),v.end(), pi(3,4), false);
+ * 	typedef Point<int> P;
+ * 	vector<P> v; v.push_back(P(4,4));
+ * 	v.push_back(P(1,2)); v.push_back(P(2,1));
+ * 	bool in = insidePolygon(v, P(3,4), false);
```
simonlindholm: does the following work?
```cpp
vector<Point<ll>> v = {{2,1}, {1,2}, {2,1}};
bool in = inPolygon(v, {3, 4}, false);
```
Either way, `int`-> `ll`, insidePolygon -> inPolygon

content/geometry/insidePolygon.h
simonlindholm: Also, can you line-wrap this description?

content/geometry/insidePolygon.h
Chillee: I did not know that was acceptable syntax.

content/geometry/insidePolygon.h
Chillee: I don't *really* understand the example already existing or the example you posted. I think it's better to have the function return 1.

I'll do
```
    vector<Point<ll>> v= {P{4,4}, P{1,2}, P{2,1}};
    bool in = inPolygon(v, P{3, 3}, false);
```

content/geometry/onSegment.h
Chillee: Dropping those parts results in `Intended for use with Point<ll>`. I think `works with integers` is a clearer way of expressing the same thing.

fuzz-tests/utilities/genPolygon.h
```diff
@@ -0,0 +1,70 @@
+#pragma once
+
+// Translated from Python code posted here: https://codeforces.com/blog/entry/63058?#comment-472788
```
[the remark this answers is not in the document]
Chillee: No to duplicates, yes to colinear.

fuzz-tests/utilities/genPolygon.h
```diff
@@ -0,0 +1,70 @@
...
+    auto pa = conquer(S[idx0], depth+1);
+    auto pb = conquer(S[!idx0], depth+1);
+    if (!pa.first || !pb.first) return {false, {}};
+    pb.second.erase(pb.second.begin());
+    pa.second.insert(pa.second.end(), all(pb.second));
+    return pa;
+}
+template<class P> vector<P> genPolygon(vector<P> pts) {
+    sort(all(pts));
+    pts.resize(unique(all(pts), [](auto a, auto b){return (a-b).dist()<EPS;}) - pts.begin());
```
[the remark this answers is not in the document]
Chillee: This is more just to eliminate duplicate points. I'll just use the default comparator.

content/geometry/insidePolygon.h
Chillee: I added some whitespace.

content/geometry/insidePolygon.h
Chillee: Line wrapped the description.

Also, why is it implicit? It's obvious if you understand that all the operations are just cross/dot products, but I think it's worth noting so you can understand at a glance.

content/geometry/insidePolygon.h
simonlindholm: It's a GCC extension, I can totally understand if you dislike it...

content/geometry/insidePolygon.h
simonlindholm: All algorithms work with integers if they don't have floating point results, and if it didn't work with integers, it would do `typedef Point<double> P;`.

Chillee: I wasn't sure that was the intended standard, since I saw that `circleTangents` was written like this:
```cpp
template<class P>
pair<P,P> circleTangents(const P &p, const P &c, double r) {
}
```
but had this in the description:

" P is intended to be Point<double>"

content/geometry/insidePolygon.h
Chillee: I think I'd rather not - I've never used it before :^)

content/geometry/insidePolygon.h
simonlindholm: Oh, we should just update circleTangents in that case. `typedef` is clearer than a description note, I think.

content/geometry/insidePolygon.h
simonlindholm: It can allow for some nice golfing, see e.g. the SCC. But yeah, let's leave it I guess.

content/geometry/onSegment.h
simonlindholm: I removed even the "Works with integers." -- I think it's pretty implicit from how most geometry algorithms operate on ints, and how the description says how to modify it for doubles.

Issue #75, opened 2019-04-25
Restructure flow algorithms


Chillee:
There's currently 3 flow implementations that (more or less) do similar things. There's HopcroftKarp, EdmondsKarp, and PushRelabel. There's also a 4th commonly used implementation (that KACTL doesn't include): Dinic's, although there is an issue for it #19 .

I've done a lot of benchmarking of flow algorithms, so let's break down these implementations.

HopcroftKarp: Only works on Bipartite Graphs, runs in `O(\sqrt{E}V)` time. Typically runs very fast (my guess is about 3x faster?)on those bipartite graphs. Decently long.

EdmondsKarp: Runs in `O(VE^2)` time. Strictly inferior to Dinic's in runtime. Fairly short.

Dinic's: Runs in `O(V^2E)` time, `O(VE log U)` with scaling. Has some special guarantees: `O(sqrt(E)*V)` on Bipartite graphs, and `O(min(sqrt(V),E^(2/3))*E)` on unit graphs. 

HLPP: Runs in `O(V^2sqrt(E))` time. Runs the fastest in practice on most graphs. Current KACTL version is missing global relabeling optimization, which makes a huge difference (~2x on Chinese flow problem, allows for `0.15` on SPOJ MATCHING).

If we look only at complexities, we see that a combination of Dinic's + HLPP dominates everything else. Provable guarantees are nice, but flow problems are also notoriously difficult to come up with worst cases for. I have a HLPP implementation that beats out my Dinic's on all the test cases I've tried it on (and also has/tied for the #1 spot on SPOJ FASTFLOW, VNSPOJ FASTFLOW, and the chinese flow problem). It also performs about equal to KACTL's Hopcroft-Karp on MATCHING. So, it seems that in practice, that HLPP implementation would probably perform the best.

Thus, I propose that we do a couple things:
1. Add global relabeling to the current HLPP implementation. This will give us the fastest possible flow implementation in general.
2. Add Dinic's. This gives us a bunch of nice provable guarantees on certain kinds of graphs, a much more understandable/traditional augmenting-paths flow algorithm, with even stronger guarantees if scaling is added.
3. With these two, we have both practical efficiency and theoretical guarantees covered, making HopcroftKarp and EdmondsKarp redundant. We can then remove them.

I think this gives us the best of all worlds.

simonlindholm:
> EdmondsKarp: Runs in O(VE^2) time. Strictly inferior to Dinic's in runtime. Fairly short.

[...]

PR #82, opened 2019-04-26
Added implementation of Dinic's

Chillee:
Comes in at 42 lines and 800 non-whitespace characters, compared to 36 lines and 592 non-whitespace characters for Edmonds-Karp, and 46 lines and 761 non-whitespace characters for Hopcroft-Karp. Note that about 4 of these lines and 100 of the characters are due to `addEdge` (whereas you need to add edges yourself for Edmonds-Karp).

Unluckily, it seems to run at `0.45` seconds on SPOJ MATCHING compared to `0.15` from Hopcroft-Karp... In addition, it doesn't return the matches themselves like Hopcroft-Karp it does.

For now I haven't included it in the document - I think that's another discussion. See #75 .

### DIFF
--- a/content/graph/Dinic.h
+++ b/content/graph/Dinic.h
@@ -7,45 +7,43 @@
  * On bipartite graphs the complexity is $O(\sqrt{V}E)$, and on unit graphs $O(\min(V^{1/2}, E^{2/3})E)).
  * Status: Tested on SPOJ FASTFLOW and SPOJ MATCHING
  */
-template<class T=int> struct Dinic {
-	const T INF = numeric_limits<T>::max();
-	struct edge {
+struct Dinic {
+	struct Edge {
 		int to, rev;
-		T c, f;
+		ll c, f;
 	};
-	vi lvl, ptr;
-	vector<vector<edge>> adj;
-	Dinic(int n) : lvl(n), ptr(n), adj(n) {}
-	void addEdge(int a, int b, T c, int rcap = 0) {
+	vi lvl, ptr, q;
+	vector<vector<Edge>> adj;
+	Dinic(int n) : lvl(n), ptr(n), q(n), adj(n) {}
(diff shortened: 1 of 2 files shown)


simonlindholm:
Looks good. Golfed to 40 lines and 678 characters. (Counted by `tr -d '[:space:]' | wc -c`; Edmonds-Karp gets 556 by the same measure.) Then ungolfed to 42 lines and 704 chars because you need to do a full blocking flow to get the right time complexity. Double-check me and I'll merge.

Chillee:
Looks good to me! I think the issue with the full blocking flow is actually important, nice catch. It seems to make a large difference on the LOJ flow problem. BTW, it is possible to get substantially better worst case complexity (`VE log U`, where U is the maximum edge capacity) with the following alteration to `calc`.

```
        for (int lim = (1 << 30); lim > 0; lim >>= 1) {
            do {
                lvl = ptr = vi(sz(q));
                int qi = 0, qe = lvl[s] = 1;
                while (qi < qe && !lvl[t]) {
                    int v = q[qi++];
                    trav(e, adj[v]) 
                        if (!lvl[e.to] && e.f < e.c && e.c - e.f >= lim) 
                            q[qe++] = e.to, lvl[e.to] = lvl[v] + 1;
                }
                while (ll p = dfs(s, t, LLONG_MAX))
                    flow += p;
            } while (lvl[t]);
        }
```
Not sure what the best way of noting that is...

Chillee:
As with our discussion, the original version is already doing a full blocking flow.

simonlindholm:
Right, I misparsed how the `ptr` stuff worked. ... also, right, perhaps merged too early, given the scaling thing. Let's do that in another PR if we decide to, I guess.

simonlindholm:
Added scaling in b552ece1f3e6f345c6853c721e637792947a6992.

Issue #79, opened 2019-04-26
Update Fuzz Tests


Chillee:
Main goals:
1. Reorganize fuzz tests to match the directory structure of `contents`.
2. Standardize tests. This involves a couple things.
a. assert when they fail/make failures obvious. Currently some tests pass by eyeballing output and determining whether it's close enough.
b. Perhaps standardize fast/slow tests, and maybe integrate with CI eventually (very stretch goal).
c. Make sure that tests are always importing their functions, as opposed to hard coding them.
3. Fix broken tests.
The following need fixing.
- [x] MinCostMaxFlow.cpp
- [x] PointInsideHull.cpp (caused by the current `typedef Point P` mess conflicting with each other).

simonlindholm:
> fast-delaunay.cpp (Not sure what's going on here... Is looks like # is test failures? That seems bad? It also floods my screen for output and takes forever. @simonlindholm ?).

`#` indicates that the test case was skipped due to duplicate points. There used to be `,` and `?` markers for skipping tests for concyclicity/colinearity reasons as well, but those were commented-out with the latest Delaunay implementation which handles that. The points of the markers are to indicate a) that the test hasn't hung, b) approximate performance, c) that the test doesn't skip all cases, in a simple-to-implement way. They could probably be removed/commented out. Real test failures terminate the test.

As for the performance, well, more tests is good, right? Delaunay triangulation probably has a lot of special cases. But a fast/slow mode for that test would likely be a good thing.

> 2sat.cpp (Running main1() causes a segfault. Don't know whether it's because it's an outdated test or 2Sat has problems).

You're just hitting the stack limit.

Chillee:
Got it. Updated the issue in response. 

`2sat.cpp` has some other issues. Namely, it's unclear to me what the tests are meant to output. For `main1`, 0 and 1 respectively. For `main2`, I can reason about the correct output. For `main3`, it looks like it should be outputting something ~95/100?

simonlindholm:
Right. You can comment out the printouts, those were mainly just for some intuition about how random 2-sat behaves. It's still `assert`ing correctness.

PR #84, opened 2019-04-26
Standardized tests for data structures

Chillee:
One note, random seeds are generally frowned upon for tests, as they make failed tests unreproducible.

Most of the DS tests are pretty good, so this is a very minor change.

See #79.


simonlindholm:
> One note, random seeds are generally frowned upon for tests, as they make failed tests unreproducible.

In my defense, that test did print its random seed at start.

I see that this moves LinkCutTree.h to data-structures/ rather than graphs/ where the code is -- that seems fine if we replace our LinkCutTree and put the new one in data-structures/.

PR #87, opened 2019-04-27
Started adding a FFTPolynomial file [WIP]

Chillee:
This PR is mainly to get a conversation started around how we should be representing our polynomials. Looking around at [MIT's implementation](https://github.com/ecnerwala/icpc-book/blob/master/content/numerical/fft.cpp) and [Adamant's implementation](https://github.com/e-maxx-eng/e-maxx-eng-aux/blob/master/src/polynomial.cpp), there seem to be 2 primary design choices.

First, how do we represent our polynomials. MIT does it with a `typedef vector<num> poly`. Adamant does it with a struct. I'm leaning towards doing it with a `vector<num>`, simply due to having slightly less boilerplate. The primary advantage of doing it in a struct is that we can then define a functions as methods of the struct instead of global properties. For example, I think doing `poly.inv().resize(K)` is more natural than `resize(inv(poly), K)`.

Second, how do we make our polynomials work across the 3 "fields" we support. Namely, doubles, ints with NTT, and ints with FFTMOD. MIT does this by wrapping int modulo operations in a struct. I think we can do this too (probably with `ModularArithmetic.h` and whatever API changes are necessary). I think with proper enough encapsulation, we should only need to change out one function, namely the `*=` operator.

I've started implementing the other operations, but I wanted to submit a PR with the basics so we could settle on a structure.


simonlindholm:
TBH I don't have a strong opinion on this. It would be nice to join this together with the existing Polynomial.h; there's probably some shared code, and it would make the code foot print of the existing code less, which is nice since it's basically never used.

Chillee:
I was more thinking that after I finish writing this we get rid of the old Polynomial class entirely. I don't think there's much point for it.

Chillee:
There's not much in the old Polynomial class, it's not documented, etc. I made a new file simply so things don't break while I write the new version.

simonlindholm:
That's fair. Not that I would mind a large PR; it gives a nice impression of how we imagine things looking in the end. Also, note that PolyRoots.h depends on Polynomial.h and will need some small changes if it's removed.

Chillee:
Ok then, if you insist :^)

Chillee:
I rebased upon the `FFTMod` PR since this code depends on that.

I made a couple changes to `ModularArithmetic.h` to make it fit better with this code.

Chillee:
I also wouldn't comment on small style things/trivial golfing for now, I'll fix them once this PR is more settled.

content/numerical/FFTPolynomial.h
```diff
@@ -0,0 +1,75 @@
...
+        B = modK(B * (poly({num(2)}) - A * B), 2 * sz(B));
+    return modK(B, sz(A));
+}
+poly &operator/=(poly &a, poly &b) {
+    if (sz(a) < sz(b))
+        return a = {};
+    int s = sz(a) - sz(b) + 1;
+    reverse(all(a)), reverse(all(b));
+    a.resize(s), b.resize(s);
+    a = a * inverse(b);
```
ecnerwala: Is there a reason you're not using `*=`, especially since `*` is defined in terms of `*=`?

content/number-theory/ModularArithmetic.h
```diff
@@ -27,4 +28,5 @@ struct Mod {
 		Mod r = *this ^ (e / 2); r = r * r;
 		return e&1 ? *this * r : r;
 	}
+	operator ll(){ return x; }
```
ecnerwala: I think it's a lot less bug-prone if you mark this conversion `explicit`.

content/numerical/FFTPolynomial.h
```diff
@@ -0,0 +1,75 @@
...
+poly deriv(poly a) {
+    if (a.empty())
+        return {};
+    poly b(sz(a) - 1);
+    rep(i, 1, sz(a)) b[i - 1] = a[i] * num(i);
+    return b;
+}
+poly integr(poly a) {
+    if (a.empty())
+        return {};
```
ecnerwala: I suspect you actually want to return `{0}` for consistency.

content/numerical/FastFourierTransform.h
```diff
@@ -27,25 +36,19 @@ void fft(vector<C> &a, vector<C> &rt, vi& rev, int n) {
...
 	if (a.empty() || b.empty()) return {};
 	vd res(sz(a) + sz(b) - 1);
 	int L = 32 - __builtin_clz(sz(res)), n = 1 << L;
-	vector<C> in(n), out(n), rt(n, 1); vi rev(n);
-	rep(i,0,n) rev[i] = (rev[i/2] | (i&1) << L) / 2;
-	for (int k = 2; k < n; k *= 2) {
-		C z[] = {1, polar(1.0, M_PI / k)};
-		rep(i,k,2*k) rt[i] = rt[i/2] * z[i&1];
-	}
+	vector<C> in(n), out(n), rt;
```
ecnerwala: Maybe make these static if you want to reuse the memory?

ecnerwala:
Oh whoops didn't see your comment. Anyways, here's some misc changes that you probably don't want to miss.

content/numerical/FFTPolynomial.h
```diff
@@ -0,0 +1,105 @@
...
+        b.resize(sz(b) * 2);
+        b *= (poly({num(1)}) + modK(a, sz(b)) - log(b));
+        b.resize(sz(b) / 2 + 1);
+    }
+    return modK(b, sz(a));
+}
+poly pow(poly a, ll m) {
+    int p = 0; int n = sz(a);
+    while (p < sz(a) && a[p].x == 0)
+        ++p;
```
ecnerwala: You need to handle the `p * m >= sz(a)` case.

Chillee:
I've finished porting all the functions (or their equivalents) from MIT, and added fuzz tests/benchmarking as well. I'd appreciate notes on what cases I should be testing in my benchmarks.

Current benchmarks look like this:
```
sub tests passed!
16ms elapsed [mine]
12ms elapsed [MIT]

add tests passed!
16ms elapsed [mine]
12ms elapsed [MIT]

div tests passed!
140ms elapsed [mine]
117ms elapsed [MIT]

mod tests passed!
238ms elapsed [mine]
206ms elapsed [MIT]

inv tests passed!
426ms elapsed [mine]
257ms elapsed [MIT]

derivative tests passed!
6ms elapsed [mine]
4ms elapsed [MIT]

integral tests passed!
13ms elapsed [mine]
7ms elapsed [MIT]

log tests passed!
524ms elapsed [mine]
454ms elapsed [MIT]

exp tests passed!
1054ms elapsed [mine]
1100ms elapsed [MIT]

pow tests passed!
52ms elapsed [mine]
36ms elapsed [mit]

eval tests passed!
557ms elapsed [mine]
263ms elapsed [MIT]

interp tests passed!
689ms elapsed [mine]
378ms elapsed [MIT]
```
Notably, `inverse`, `eval`, and `interp` show significant performance differences. I'll be taking a look at those next.

We'll probably also want to split these functions up into files.

ecnerwala:
The performance differences probably all arise from doing twice as many fft's in inverse.

ecnerwala:
Do you handle `pow(*, 0)` correctly?

Chillee:
> Do you handle pow(*, 0) correctly?

I believe so. 

> The performance differences probably all arise from doing twice as many fft's in inverse.

Ok, so I moved some stuff around (namely, in the benchmarks where I was regenerating vectors constantly). I'll update my comment with the new benchmarks. Notably, it seems like the current `inverse` is only about 40% slower - not 100%. 

So the disparity in runtimes for `interp/eval` are the biggest differences now.

ecnerwala:
You may want to document various assumptions later. Off the top of my head:

- `inverse` expects `a[0] != 0`
- `log` expects `a[0] == 1`
- `exp` expects `a[0] == 0`
- `pow({0}, 0) == {0}`

Chillee:
> Document various assumptions

Yeah that's definitely that should be done. Splitting these into multiple files should luckily give us plenty of space for documentation about requirements and such.

Chillee:
Seems like the main cause of the relatively slow performance is the big performance discrepancy in `inverse` for small values. For large values the difference is only about 40%, but values of size 10 it's around 200%.

Chillee:
Now that we've reached what I consider reasonable feature/performance parity, I'll start formatting things into files/writing some documentation next.

ecnerwala:
I suspect you can get some reasonable gains on small cases by resuing the `rt` and `rev` arrays (store globally). I'm not a huge fan of having `naive_mul` both on principle and because it's a bunch of lines, but idk.

Chillee:
So the current `FFTMod` PR has some work on reusing the `rt` array - I think the `rev` array needs to be recomputed for each different size.

It seems that for arrays around size 10, there's about a 50% gain in doing so (so I don't think it'll close the gap).

I do agree that it's kind of an ugly hack :'(. Depending on how much we think performance matters for `eval`/`interp`, we might not even want to add it. However, a 2x performance gain is definitely something to be considered.

I also think that doing `naive_mul` isn't too bad. It adds about 6 lines total:
```cpp
poly &operator*=(poly &a, const poly &b) {
    if (sz(a) + sz(b) < 100){
        poly res(sz(a) + sz(b) - 1);
		rep(i,0,sz(a)) rep(j,0,sz(b))
			res[i + j] = (res[i + j] + a[i] * b[j]);
        return (a = res);
    }
    return a = conv(a, b);
}
```

ecnerwala:
You can reuse the rev array with some bitshifting, see the MIT implementation (https://github.com/ecnerwala/icpc-book/blob/269f583e2b2a8497caac1f488ce3e7dac2f52ba3/content/numerical/fft.cpp#L67).

Chillee:
It seems to make no impact on performance (perhaps even a negative impact).

Chillee:
I've reorganized the functions into 5 files (with their dependencies by the side (including transitive dependencies): 

1. PolynomialBase
1. PolynomialMod [1]
1. PolynomialPow [1]
1. PolynomialEval [1], [2]
1. PolynomialInterpolate [1], [2], [3], [4] (wouldn't need [3] other than `deriv`)

content/numerical/PolynomialBase.h
```diff
@@ -0,0 +1,57 @@
...
+	poly c = a;
+	trav(i, c) i = i * b;
+	return c;
+}
+#define OP(o, oe) \
+	poly operator o(poly a, poly b) { \
+		poly c = a; \
+		return c oe b; \
+	}
+OP(*, *=) OP(+, +=) OP(-, -=);
```
ecnerwala: You can get rid of the `oe` parameter, just use `o##=` which concatenates tokens.

```
#define OP(o) \
	poly operator o(poly a, poly b) { \
		poly c = a; \
		return c o##= b; \
	}
OP(+) OP(-) OP(*)
```

content/numerical/PolynomialBase.h
```diff
@@ -0,0 +1,57 @@
...
+#include "FastFourierTransformMod.h"
+// #include "NumberTheoreticTransform.h"
+
+typedef Mod num;
+typedef vector<num> poly;
+vector<Mod> conv(vector<Mod> a, vector<Mod> b) {
+	auto res = convMod<mod>(vl(all(a)), vl(all(b)));
+	// auto res = conv(vl(all(a)), vl(all(b)));
+	return vector<Mod>(all(res));
+}
```
ecnerwala: Is this layer of wrapping useful? Why not put it directly in `operator *=`?

content/numerical/PolynomialBase.h
```diff
@@ -0,0 +1,57 @@
...
+ * Author: chilli, Andrew He, Adamant
+ * Date: 2019-04-27
+ * Description: A FFT based Polynomial class.
+ */
+#pragma once
+
+#include "../number-theory/ModularArithmetic.h"
+#include "FastFourierTransform.h"
+#include "FastFourierTransformMod.h"
+// #include "NumberTheoreticTransform.h"
```
ecnerwala: Nit: NTT will be the common case, if that matters to you?

content/numerical/PolynomialPow.h
```diff
@@ -0,0 +1,45 @@
...
+		return {};
+	poly b(sz(a) - 1);
+	rep(i, 1, sz(a)) b[i - 1] = a[i] * num(i);
+	return b;
+}
+poly integr(poly a) {
+	if (a.empty()) return {0};
+	poly b(sz(a) + 1);
+	b[1] = num(1);
+	rep(i, 2, sz(b)) b[i] = b[mod%i]*Mod(-mod/i+mod);
```
ecnerwala: Doesn't work for complex, probably worth noting.

content/numerical/PolynomialBase.h
Chillee: ??? C macros astound me again!

Neat trick.

content/numerical/FastFourierTransform.h
```diff
@@ -3,22 +3,31 @@
...
+	vi rev(n);
+	rep(i,0,n) rev[i] = (rev[i / 2] | (i & 1) << L) / 2;
+	if (rt.empty()) {
+		rt.assign(n, 1);
+		for (int k = 2; k < n; k *= 2) {
+			Cd z[] = {1, polar(1.0, M_PI / k)};
+			rep(i, k, 2 * k) rt[i] = Cd(rt[i / 2]) * z[i & 1];
+		}
+	}
 	rep(i,0,n) if (i < rev[i]) swap(a[i], a[rev[i]]);
```
ecnerwala: You can merge this loop with the generating loop if you want, would keep things more compact and (marginally) improve cache locality.

content/numerical/PolynomialBase.h
Chillee: Good point - I was originally testing out a couple different ideas to encapsulate `FFT`, and wasn't sure what'd be necessary.

content/numerical/PolynomialBase.h
Chillee: This is currently commented out for a stupid reason -  `NumberTheoreticTransform.h` doesn't compile properly on its own (due to redefinitions of const mod). It is good to know that NTT is the common case though.

content/numerical/PolynomialPow.h
Chillee: Is it even worth noting stuff that doesn't work with complex numbers? Or is it reasonable to assume that every problem involving polynomials/series will be done on a finite field?

ecnerwala: Idk, the field of reals seems like it could be useful sometimes.

content/numerical/PolyInterpolate.h
```diff
@@ -1,25 +1,24 @@
 /**
- * Author: Simon Lindholm
- * Date: 2017-05-10
- * License: CC0
- * Source: Wikipedia
+ * Author: chilli, Andrew He, Adamant
```
ecnerwala: Much of my work comes from THU, probably should give them credit too.

content/number-theory/ModularArithmetic.h
```diff
@@ -12,21 +12,19 @@
...
-		Mod r = *this ^ (e / 2); r = r * r;
-		return e&1 ? *this * r : r;
+        ll ans = 1, b = (*this).v;
+        for (; e; b = b * b % mod, e /= 2)
+            if (e & 1) ans = ans * b % mod;
+        return ans;
 	}
-	explicit operator ll() const { return x; }
-};
+    explicit operator ll() const { return v; }
```
ecnerwala: Uh, tabs?

Chillee: :'(

Chillee:
I removed the `euclid` dependency in `ModularArithmetic.h` for 2 reasons:

1. It makes the total Polynomial code longer :^)
2. I suspect that there's not much point in supporting division on coprime values. I can't think of places where one might want to use `ModularArithmetic` on a ring that isn't a finite field. If someone really needs to, it's easy enough to change it.

ecnerwala:
If you want, here's my modinv:
```int modinv(int a, int m) { return a == 1 ? 1 : m - 1ll * modinv(m % a, a) * m / a; }```
Requires `1 <= a < m` and `gcd(a,m) == 1`

ecnerwala:
Also, empirically, modinv is somewhere between 1x and 2x faster than modular exponentiation.

PR #89, opened 2019-05-01
Updated policy hashmap with custom hash

### LINKED ISSUE #86: Add custom hash function for `gp_hash_table`
https://github.com/ecnerwala/icpc-book/blob/master/content/various/Hashmap.h

Also see my blog post: https://codeforces.com/blog/entry/60737

Chillee:
See https://ideone.com/4frDuM.

```
8ms elapsed [custom hash insert]
2ms elapsed [custom hash read]
2847ms elapsed [no hash insert]
1701ms elapsed [no hash read]
```

It's a worst case example, of course, but it's a fairly natural one. I insert `v, v<<1, v<<2, ... v<<40`. We see a >300x slowdown for both insertions and reads (8/2ms to 2847/1701ms respectively).

I also have some discussion about it on the original post under "A better hash function".

I also don't think the custom preallocation is a big deal. Maybe it's worth including since it's not that much more, but my benchmarks don't show much of an improvement (maybe 5-10%).

Fixes #86 .

### DIFF
--- a/content/data-structures/HashMap.h
+++ b/content/data-structures/HashMap.h
@@ -9,7 +9,16 @@
 #pragma once
 
 #include <bits/extc++.h> /** keep-include */
-__gnu_pbds::gp_hash_table<ll, int> h({},{},{},{}, {1 << 16});
+typedef unsigned long long ull;
+struct chash {
+    ull hash_f(ull x) {
+        x += 0x9e3779b97f4a7c15;
+        x = (x ^ (x >> 30)) * 0xbf58476d1ce4e5b9;
+        x = (x ^ (x >> 27)) * 0x94d049bb133111eb;
+        return x ^ (x >> 31);
+    }
+};
+__gnu_pbds::gp_hash_table<ll, int, chash> h({},{},{},{}, {1 << 16});
 
 /** For CodeForces, or other places where hacking might be a problem:
 const int RANDOM = chrono::high_resolution_clock::now().time_since_epoch().count();
--- a/content/data-structures/HashMap.h
+++ b/content/data-structures/HashMap.h
@@ -9,16 +9,19 @@
 #pragma once
 
 #include <bits/extc++.h> /** keep-include */
-struct chash { // To use most bits rather than just the lowest ones:
-    const uint64_t C = ll(2e18 * M_PI) + 71; // large odd number
-    ll operator()(ll x) const { return __builtin_bswap64(x*C); }
+// To use most bits rather than just the lowest ones:
+struct chash {
+	const uint64_t C = ll(2e18 * M_PI) + 71; // large odd number
+	ll operator()(ll x) const { return __builtin_bswap64(x*C); }
 };
-__gnu_pbds::gp_hash_table<ll, int, chash> h({},{},{},{}, {1 << 16});
+__gnu_pbds::gp_hash_table<ll,int,chash> h({},{},{},{},{1<<16});
 
 /** For CodeForces, or other places where hacking might be a problem:
+
 const int RANDOM = chrono::high_resolution_clock::now().time_since_epoch().count();
-struct chash {
-    int operator()(ll x) const { return x + RANDOM; }
+struct chash { // To use most bits rather than just the lowest ones:
+	const uint64_t C = ll(2e18 * M_PI) + 71; // large odd number
+	ll operator()(ll x) const { return __builtin_bswap64((x^RANDOM)*C); }
 };
 __gnu_pbds::gp_hash_table<ll, int, chash> h({},{},{},{}, {1 << 16});
 */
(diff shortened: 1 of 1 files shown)


simonlindholm:
Ugh, so `gp_hash_table` uses the *lower order* bits by default? Why... And then it has a mod implementation, which doesn't even do precomputed Barrett reduction. I thought this hash table was supposed to be good...

I agree that fixing this is sensible (perhaps with a comment about how it behaves if you don't). Your implementation is rather long, though, and it seems slow which is at odds with needing a really fast hash table (else you'd be using `unordered_map`). I wonder if we could get away with e.g. `__builtin_bswap64(x * <random number>ULL)`. Or we could declare our own range_hashing which multiplies by a large number and then shifts rather than masks (Mozilla's hash table does this, with the large number equal to round_to_odd(2^64 * phi)). Not sure how verbose that would be.

>I also don't think the custom preallocation is a big deal. Maybe it's worth including since it's not that much more, but my benchmarks don't show much of an improvement (maybe 5-10%).

Yep, that matches my benchmarks as well. I figure that if you really need fast hashing, 10% can make a difference.

simonlindholm:
> Or we could declare our own range_hashing which multiplies by a large number and then shifts rather than masks (Mozilla's hash table does this, with the large number equal to round_to_odd(2^64 * phi)). Not sure how verbose that would be.

Turns out to be fairly long:

```cpp
using namespace __gnu_pbds;
struct reducer {
	typedef size_t size_type;
	size_t sh;
	void notify_resized(size_t s) { sh = 1 + __builtin_clzll(s); }
	size_t operator()(size_t s) const {
		return (s * 0x517cc1b727220a95ULL) >> sh;
	}
};

typedef __gnu_pbds::gp_hash_table<ll, int, hash<ll>, equal_to<ll>, reducer,
    linear_probe_fn<>, hash_standard_resize_policy<hash_exponential_size_policy<>,
        hash_load_check_resize_trigger<>>> ht;
```

Also, it's used twice for each key: first `h = reduce(hash(key))`, then `reduce(h)`, `reduce(h+1)`, ...  Which is semi-reasonable when reducing just does `% size` (though wasteful), but with the custom reducer it destroys cache locality...

So I think modifying the hash function is the way to go.

simonlindholm:
```cpp
struct chash {
    ll operator()(ll x) const {
        return __builtin_bswap64(x * 0x517cc1b727220a95ULL);
    }
};
```
seems to work decently well: https://ideone.com/irwNSt
It will use all bits as long as the size is at least 256. Need to think about how to golf the annoying constant.

simonlindholm:
Seems like basically any large enough odd constant will do, preferably with a nice mix of ones/zeroes.  `size_t(2e18 * M_PI) + 99` works if size_t = ull, but probably breaks silently for 32-bit.

simonlindholm:
Here's a concrete suggestion:

```cpp
// To use most bits rather than just the lowest ones:
struct chash {
    const uint64_t C = ll(2e18 * M_PI) + 71; // large odd number
    ll operator()(ll x) const { return __builtin_bswap64(x*C); }
};

__gnu_pbds::gp_hash_table<ll, int, chash> h{{},{},{},{}, {1<<16}};
```

bswap isn't perfect -- high bits in the upper half of bytes can still end up ignored, and you can construct bad cases like if you store all multiplies of 2^52 and nothing else you'll have problems, especially given linear probing. But it might be good enough?

simonlindholm:
Also, this makes me wonder how good the suggestion of `return x ^ RANDOM;` for Codeforces can be. That should affect all elements equally...

Chillee:
Hmm... Yeah I suspect you need to do `x+RANDOM` or `x*RANDOM`.

I think the suggestion is good enough. The most common accidental hash collision by far one is likely to run into is the example I listed.

content/data-structures/HashMap.h
```diff
@@ -9,12 +9,16 @@
...
+    const uint64_t C = ll(2e18 * M_PI) + 71; // large odd number
+    ll operator()(ll x) const { return __builtin_bswap64(x*C); }
+};
+__gnu_pbds::gp_hash_table<ll, int, chash> h({},{},{},{}, {1 << 16});
 
 /** For CodeForces, or other places where hacking might be a problem:
 const int RANDOM = chrono::high_resolution_clock::now().time_since_epoch().count();
 struct chash {
-    int operator()(ll x) const { return x ^ RANDOM; }
+    int operator()(ll x) const { return x + RANDOM; }
```
simonlindholm: `+ RANDOM` and `^ RANDOM` are about equally bad. `__builtin_bswap64((x ^ RANDOM) * C);` would be alright.

Chillee: Err.. that was what I was imagining I suppose :)

PR #91, opened 2019-05-04
Renamed to OnSegment.h

Chillee:
To align with modern KACTL naming conventions.


simonlindholm:
Only problem with this is that it makes the header look like this:
image
:)

Shortening some other filename seems like the simplest solution. We could also decrease the font size when a certain number of characters is hit - the header tex code is generated by Python (print_header), so this wouldn't be too hard.

Chillee:
I'm gonna close this PR since I don't think I can work on this - my latex compilation is strange and gives different results from what you're using.

I'll open an issue though.

PR #93, opened 2019-05-05
Updated CircleIntersection.h

Chillee:
3 main changes. I didn't understand why any of these existed in the first place so if there's a reason, I'll revert that change from this PR: 
1. Changed the API to return 0/1/2 depending on the number of intersection points. Currently, the function doesn't distinguish between 1 and 2 return points. I suppose this is the equivalent of testing for `strict` in functions like `inPolygon`.
2. Changed `out` from a pointer to a reference.
3. Removed the `d2 > r*r` check.  Is this check here for higher precision?

### DIFF
--- a/content/geometry/CircleIntersection.h
+++ b/content/geometry/CircleIntersection.h
@@ -1,9 +1,8 @@
 /**
- * Author: Simon Lindholm, Victor Lecomte, chilli
+ * Author: Simon Lindholm
  * Date: 2015-09-01
  * License: CC0
- * Source: https://vlecomte.github.io/cp-geo.pdf
- * Description: Computes a pair of points at which two circles intersect. Returns false in case of no intersection.
+ * Description: Computes the pair of points at which two circles intersect. Returns false in case of no intersection.
  * Status: somewhat tested
  */
 #pragma once
@@ -11,13 +10,13 @@
 #include "Point.h"
 
 typedef Point<double> P;
-int circleInter(P a,P b,double r1,double r2,pair<P, P>* out){
+bool circleInter(P a,P b,double r1,double r2,pair<P, P>* out) {
 	if (a == b) { assert(r1 != r2); return false; }
-	P delta = b - a;
-	double d2 = delta.dist2(), p = (d2 + r1*r1 - r2*r2)/(d2*2);
-	double h2 = r1*r1 - p*p*d2;
-	if (h2 < 0) return false;
-	P mid = a + delta*p, per = delta.perp() * sqrt(h2 / d2);
+	P vec = b - a;
+	double d2 = vec.dist2(), sum = r1+r2, dif = r1-r2,
+	       p = (d2 + r1*r1 - r2*r2)/(d2*2), h2 = r1*r1 - p*p*d2;
+	if (sum*sum < d2 || dif*dif > d2) return false;
+	P mid = a + vec*p, per = vec.perp() * sqrt(fmax(0, h2) / d2);
 	*out = {mid + per, mid - per};
 	return true;
 }
--- a/content/geometry/CircleIntersection.h
+++ b/content/geometry/CircleIntersection.h
@@ -1,8 +1,8 @@
 /**
- * Author: Simon Lindholm
+ * Author: Simon Lindholm, Victor Lecomte, chilli
  * Date: 2015-09-01
  * License: CC0
- * Description: Computes a pair of points at which two circles intersect. Returns false in case of no intersection.
+ * Description: Computes a pair of points at which two circles intersect. Returns number of intersections. If intersection is only at one point then that point is returned twice.
  * Status: somewhat tested
  */
 #pragma once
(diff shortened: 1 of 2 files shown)


simonlindholm:
> Changed the API to return 0/1/2 depending on the number of intersection points.

Makes sense, I suppose, not that you'd ever special-case the number of intersections apart from sometimes wanting "strict" behavior.

> Changed out from a pointer to a reference.

I have a mild preference for pointers for out-params in general, because it makes the call-site clearer.

> Removed the d2 > r*r check. Is this check here for higher precision?

No, that check was just sloppiness on my part, thanks.

content/geometry/CircleIntersection.h
```diff
@@ -1,25 +1,22 @@
...
 typedef Point<double> P;
-bool circleIntersection(P a, P b, double r1, double r2,
-		pair<P, P>* out) {
-	P delta = b - a;
-	assert(delta.x || delta.y || r1 != r2);
-	if (!delta.x && !delta.y) return false;
-	double r = r1 + r2, d2 = delta.dist2();
+int circleInter(P a,P b,double r1,double r2,pair<P, P>& out){
+	P delta = b - a; double d2 = delta.dist2();
+	if (d2 == 0) { assert(r1 != r2); return 0; }
```
simonlindholm: `if (a == b) { ... }` at the top of the function?

content/geometry/CircleIntersection.h
```diff
@@ -1,25 +1,22 @@
...
 
 typedef Point<double> P;
-bool circleIntersection(P a, P b, double r1, double r2,
-		pair<P, P>* out) {
-	P delta = b - a;
-	assert(delta.x || delta.y || r1 != r2);
-	if (!delta.x && !delta.y) return false;
-	double r = r1 + r2, d2 = delta.dist2();
+int circleInter(P a,P b,double r1,double r2,pair<P, P>& out){
+	P delta = b - a; double d2 = delta.dist2();
```
simonlindholm: move this down, possibly eliminate delta or reuse b for it; declarations of different types on the same line is a bit ugly, and the double declarations could be combined

content/geometry/CircleIntersection.h
```diff
@@ -1,25 +1,22 @@
...
+	if (d2 == 0) { assert(r1 != r2); return 0; }
 	double p = (d2 + r1*r1 - r2*r2) / (2.0 * d2);
 	double h2 = r1*r1 - p*p*d2;
-	if (d2 > r*r || h2 < 0) return false;
+	if (h2 < 0) return 0;
 	P mid = a + delta*p, per = delta.perp() * sqrt(h2 / d2);
-	*out = {mid + per, mid - per};
-	return true;
+	out = {mid + per, mid - per};
+	return 1 + sgn(h2);
```
simonlindholm: `1 + (h2 > 0)`

simonlindholm:
> >   Removed the d2 > r*r check. Is this check here for higher precision?
>
> No, that check was just sloppiness on my part, thanks.

Well, actually, hm. If we do want to provide a "strict" flag, we should do it properly, it makes no sense to do it based on double-rounded results, and then this sort of check would come into play (but you'd need some additional one for circle within circle as well, or expand the math involving sqrts/divisions into a single integer check). Or we could just make it return bool again.

content/geometry/CircleIntersection.h
```diff
@@ -10,15 +11,12 @@
...
+	if (a == b) { assert(r1 != r2); return false; }
 	P delta = b - a;
-	assert(delta.x || delta.y || r1 != r2);
-	if (!delta.x && !delta.y) return false;
-	double r = r1 + r2, d2 = delta.dist2();
-	double p = (d2 + r1*r1 - r2*r2) / (2.0 * d2);
+	double d2 = delta.dist2(), p = (d2 + r1*r1 - r2*r2)/(d2*2);
 	double h2 = r1*r1 - p*p*d2;
-	if (d2 > r*r || h2 < 0) return false;
+	if (h2 < 0) return false;
```
simonlindholm: Hm, so. This does lose some numeric precision; now circles that are disjoint might get an intersection due to bad rounding if we're not favored by the floating point gods. (We had similar problems in other cases even before.)

If I did my math right, a numerically stable way of writing this condition would be `(a+b+c)*(a+b-c)*(b-a+c)*(a-b+c) < 0`, where `a = r1`, `b = r2`, `c = sqrt(d2)` (of which the first term could be eliminated, and the rest probably reformulated in terms of inequalities). I wonder if the rest of the function could be written in terms of those variables...

simonlindholm: Well, I guess we could keep the rest of the function, but I think returning false if either `(r1 + r2)^2 < d2` or `(r1 - r2)^2 > d2`, and then letting `h2 = max(0.0, current expr);`, would be worthwhile

simonlindholm:
Merged with a bunch of changes (the condition for returning false changed to `sum*sum < d2 || dif*dif > d2` being perhaps the main one).

PR #96, opened 2019-05-05
Change x.begin() to begin(x)

Chillee:
`begin(x)` is a strict superset of what `x.begin()` can do.

See https://stackoverflow.com/questions/26290316/difference-between-vectorbegin-and-stdbegin

I realize that KACTL almost never uses C-style arrays, but we also save 2 characters from this :^)


simonlindholm:
Sounds good. Only reason I could see not to is if there are local `begin`/`end` variables that shadow the functions.

PR #97, opened 2019-05-05
Updated Convex Hull

Chillee:
The new convexHull is also ~10% faster: https://ideone.com/FxPMuy

Submission on kattis `convexHull`: https://ideone.com/1Pfk6H

Submission on kattis `roberthood`: https://ideone.com/a97DIZ


fuzz-tests/geometry/ConvexHull.cpp
```diff
@@ -0,0 +1,60 @@
...
+    timeit(string label = "???") : label(label) { begin = chrono::high_resolution_clock::now(); }
+    ~timeit() {
+        auto end = chrono::high_resolution_clock::now();
+        auto duration = chrono::duration_cast<chrono::milliseconds>(end - begin).count();
+        cerr << duration << "ms elapsed [" << label << "]" << endl;
+    }
+};
+int main() {
+    const int SZ = 1e4;
+    rep(t,0,1000) {
```
simonlindholm: `1000` is super low, and `1e3`/`1e4` is super high

fuzz-tests/geometry/ConvexHull.cpp
```diff
@@ -0,0 +1,60 @@
...
+        const int GRID=1e3;
+        vector<P> pts(SZ);
+        rep(i,0,SZ) pts[i] = P(rand()%GRID, rand()%GRID);
+        auto res = convexHull(pts);
+        auto res2 = old::convexHull(pts);
+        rep(i,0,sz(res2)) {
+            assert(pts[res2[i]] == res[i]);
+        }
```
simonlindholm: `assert(sz(res) == sz(res2))` as well

content/geometry/ConvexHull.h
```diff
@@ -24,22 +24,17 @@ Points on the edge of the hull between two other points are not considered part
...
-	if (S[u[0]] == S[u[1]]) return {0};
-	l.insert(l.end(), u.rbegin()+1, u.rend()-1);
-	return l;
-}
+vector<P> convexHull(vector<P> pts) {
+	if (sz(pts) <= 1) return pts;
+	sort(all(pts));
+	vector<P> h(sz(pts)+1);
+	int t = 0;
+	for (int it = 0, s = t; it < 2; it++, s = t) {
```
simonlindholm: `s = --t` at the end?

content/geometry/ConvexHull.h
```diff
@@ -24,22 +24,17 @@ Points on the edge of the hull between two other points are not considered part
...
+		trav(p, pts){
+			while (t >= s + 2 && h[t-2].cross(h[t-1], p) <= 0) t--;
```
simonlindholm: if we decrease `t` by `1` everywhere, that would shave off a few characters. not sure if worth the code clarity loss

Chillee: I'm not actually even sure that's true... the only place where we could save some characters is `h[t-1]` -> `h[t]`, but I think that might be offset elsewhere (like at the end, for example).

Chillee:
Oops I didn't mean to check in the `lineHullIntersection`, I meant to check in a different file.

fuzz-tests/geometry/ConvexHull.cpp
Chillee: Upped to 1e5 iterations and 1e2 amount of points. 

One thing to note is that if we generate `N` random points, the convex hull will contain ~ `log(N)` points.

content/geometry/ConvexHull.h
Chillee: :+1: 

I can also fit in the `reverse` in the loop.

content/geometry/ConvexHull.h
simonlindholm: there's also `t >= s + 2` -> `t > s`, and the end becomes an addition instead of a subtraction

content/geometry/ConvexHull.h
```diff
@@ -1,45 +1,36 @@
...
- * Source: Basic algorithm knowledge
+ * Author: Stjepan Glavina, chilli
+ * Date: 2019-05-05
+ * License: Unlicense
+ * Source: https://github.com/stjepang/snippets/blob/master/convex_hull.cpp
  * Description:
 \\\begin{minipage}{75mm}
 Returns a vector of indices of the convex hull in counter-clockwise order.
 Points on the edge of the hull between two other points are not considered part of the hull.
+Works with integers.
```
simonlindholm: This is pretty obvious, the question is whether it works with doubles. I would imagine it does?

content/geometry/HullDiameter.h
```diff
@@ -0,0 +1,22 @@
+/**
+ * Author: Oleksandr Bacherikov, chilli
+ * Date: 2019-05-05
+ * License: CC0
+ * Source: https://codeforces.com/blog/entry/48868
+ * Description: Returns the two points with max distance on a convex hull. If input isn't a convex hull, call "convexHull" first.
```
simonlindholm: are colinear points allowed? duplicate points?

content/geometry/HullDiameter.h
```diff
@@ -0,0 +1,22 @@
...
+ * Works with integers.
+ * Status: Fuzz-tested, submitted on kattis roberthood
+ */
+#pragma once
+
+array<P, 2> hullDiameter(vector<P> S) {
+	int n = sz(S), j = n < 2 ? 0 : 1;
+	pair<ll, array<P, 2>> res;
```
simonlindholm: this needs to be initialized to something

fuzz-tests/geometry/ConvexHull.cpp
simonlindholm: I think we want to test small grids, though -- that has a higher chance of triggering degenerate cases. You could make N equal `(int)exp(rand() % something)`, and similar with grid size.

content/geometry/ConvexHull.h
Chillee: Shouldn't geometry algorithms that work on integers be a superset of those that work on doubles? It's obvious if you understand what the algorithm is doing/what operations it's using, but I think it's good to make clear.

content/geometry/HullDiameter.h
Chillee: https://stackoverflow.com/questions/9025792/does-the-default-constructor-of-stdpair-set-basic-types-int-etc-to-zero

content/geometry/HullDiameter.h
Chillee: They aren't allowed - added a note in the description.

content/geometry/ConvexHull.h
Chillee: The only possible source of numerical error is the operation `h[t-2].cross(h[t-1], p) <= 0`. And that's as good as it gets.

Chillee: Also, I removed that from the description.

content/geometry/HullDiameter.h
simonlindholm: oh, interesting. might still want to return `{S[0], S[0]}` if `sz(S) == 1` though... or assert `sz(S) >= 2`

content/geometry/HullDiameter.h
```diff
@@ -0,0 +1,23 @@
+/**
+ * Author: Oleksandr Bacherikov, chilli
+ * Date: 2019-05-05
+ * License: CC0
+ * Source: https://codeforces.com/blog/entry/48868
+ * Description: Returns the two points with max distance on a convex hull (no duplicate/collinear points).
```
simonlindholm: I guess it might need to be CCW as well

content/geometry/HullDiameter.h
```diff
@@ -3,15 +3,17 @@
...
+ * no duplicate/collinear points). If input isn't a convex hull, call
+ * "convexHull" first.
  * Status: Fuzz-tested, submitted on kattis roberthood
  */
 #pragma once
 
 typedef Point<ll> P;
 array<P, 2> hullDiameter(vector<P> S) {
 	int n = sz(S), j = n < 2 ? 0 : 1;
+	assert(n>=2);
```
simonlindholm: space around `>=`, and you can set `j = 1` unconditionally given this

Chillee: Actually, I'm back to thinking that it makes marginally more sense if `{S[0], S{0]}` is returned.

PR #98, opened 2019-05-05
Unified bipartite matching APIs

Chillee:
Motivated by someone asking me how they could use HopcroftKarp to do MinimumVertexCover. Admittedly, they didn't notice MinimumVertexCover at all, but swapping MinimumVertexCover to use HopcroftKarp requires more than a single line change for non reason at all.

I also added some more documentation for HopcroftKarp (primarily copying from DfsMatching).


content/graph/DFSMatching.h
```diff
@@ -5,37 +5,35 @@
  * Source:
  * Description: This is a simple matching algorithm but should
  * be just fine in most cases. Graph $g$ should be a list of
- * neighbours of the left partition. $n$ is the size of the left
- * partition and $m$ is the size of the right partition.
- * If you want to get the matched pairs, $match[i]$ contains
+ * neighbours of the left partition, and $btoa$ should be
+ * a vector with size equal to the size of the right partition.
```
simonlindholm: "full of -1's"

content/graph/DFSMatching.h
```diff
@@ -5,37 +5,35 @@
...
 		}
 	return 0;
 }
-int dfs_matching(const vector<vi>& g, int n, int m) {
-	match.assign(m, -1);
-	rep(i,0,n) {
-		seen.assign(m, 0);
+int dfsMatching(vector<vi> g, vi& btoa) {
+	rep(i,0,sz(g)) {
+		vector<bool> vis(sz(btoa));
```
simonlindholm: making this global again would remove an allocation here, and is likely shorter as well

content/graph/DFSMatching.h
```diff
@@ -5,37 +5,35 @@
  * Source:
  * Description: This is a simple matching algorithm but should
  * be just fine in most cases. Graph $g$ should be a list of
- * neighbours of the left partition. $n$ is the size of the left
- * partition and $m$ is the size of the right partition.
- * If you want to get the matched pairs, $match[i]$ contains
+ * neighbours of the left partition, and $btoa$ should be
+ * a vector with size equal to the size of the right partition.
+ * If you want to get the matched pairs, $btoa[i]$ contains
```
simonlindholm: This wording is fluffy, let's rewrite it as "Returns the size of the matching. $btoa[i]$ will contain the match ..."

content/graph/DFSMatching.h
```diff
@@ -5,37 +5,35 @@
  * Source:
  * Description: This is a simple matching algorithm but should
  * be just fine in most cases. Graph $g$ should be a list of
```
simonlindholm: Let's shorten the intro to "Simple bipartite matching algorithm."

content/graph/hopcroftKarp.h
```diff
@@ -3,7 +3,11 @@
  * Date: 2009-10-13
  * License: CC0
  * Source: N/A
- * Description: Find a maximum matching in a bipartite graph.
+ * Description: Find a maximum matching in a bipartite graph. Graph $g$ should
```
simonlindholm: maybe replace the intro sentence by "Faster bipartite matching algorithm."?

content/graph/DFSMatching.h
Chillee: I don't think it saves an allocation - we still need to do `vis.assign(...)`. Main advantage would be that we don't need to pass it through the parameters.

PR #100, opened 2019-05-05
Rewrote `PointInsideHull`

Chillee:
This is the more standard way to write `PointInsideHull.h`. I cited where I first learned of this method, but I'm pretty sure that it's the standard.

Shortened it by 9 lines.


content/geometry/PointInsideHull.h
```diff
@@ -1,13 +1,13 @@
...
- * Date: 2017-04-13
+ * Author: chilli
+ * Date: 2019-05-17
  * License: CC0
- * Source: Inspired by old, broken tinyKACTL
- * Description: Determine whether a point t lies inside a given polygon (counter-clockwise order).
- * The polygon must be such that every point on the circumference is visible from the first point in the vector.
- * It returns 0 for points outside, 1 for points on the circumference, and 2 for points inside.
+ * Source: https://github.com/ngthanhtrung23/ACM_Notebook_new
+ * Description: Determine whether a point t lies inside a convex hull. Returns
```
simonlindholm: adding "(counter-clockwise order, with no colinear points)" would be useful, if that's the case

simonlindholm: can also shorten the strictness comment slightly into "if strict is true, points on the boundary are not included" or some such

content/geometry/PointInsideHull.h
```diff
@@ -17,24 +17,15 @@
...
+bool inHull(const vector<P> &l, P p, bool strict = true) {
+    int a = 1, b = l.size() - 1, r = !strict;
+	if (sz(l) < 3) return r && onSegment(l[0], l.back(), p);
+    if (sideOf(l[0], l[a], l[b]) > 0) swap(a, b);
+    if (sideOf(l[0], l[a], p) >= r || sideOf(l[0], l[b], p) <= -r) return false;
+    while (abs(a - b) > 1) {
+        int c = (a + b) / 2;
+        (sideOf(l[0], l[c], p) > 0 ? b : a) = c;
+    }
+    return !(sgn(l[a].cross(l[b], p)) >= r);
```
simonlindholm: `sgn(...) < r`

content/geometry/sideOf.h
```diff
@@ -13,10 +13,8 @@
 #include "Point.h"
 
 template<class P>
-int sideOf(const P& s, const P& e, const P& p) {
-	auto a = (e-s).cross(p-s);
-	return (a > 0) - (a < 0);
-}
+int sideOf(P s, P e, P p) { return sgn((e-s).cross(p-s)); }
```
simonlindholm: `sgn(s.cross(e, p))`

fuzz-tests/geometry/PointInsideHull.cpp
```diff
@@ -11,22 +11,31 @@ typedef pair<int, int> pii;
...
+ostream &operator<<(ostream &os, P p) { return cout << "(" << p.x << "," << p.y << ")"; }
 
 int main() {
 	rep(it,0,100000) {
 		int N = rand() % 15;
 		vector<P> ps, ps2;
 		rep(i,0,N) ps.emplace_back(rand() % 20 - 10, rand() % 20 - 10);
 		trav(i, convexHull(ps)) ps2.push_back(ps[i]);
-		if (ps2.empty()) continue;
+		if (sz(ps2) <= 0) continue;
```
simonlindholm: FWIW I'd say that `ps2.empty()` is clearer

content/geometry/PointInsideHull.h
```diff
@@ -17,24 +17,15 @@
...
-	if (sideOf(H[0], H[mid], p) >= 0)
-		return insideHull2(H, mid, R, p);
-	return insideHull2(H, L, mid+1, p);
-}
 
-int insideHull(const vector<P>& hull, const P& p) {
-	if (sz(hull) < 3) return onSegment(hull[0], hull.back(), p);
-	else return insideHull2(hull, 1, sz(hull), p);
+bool inHull(const vector<P> &l, P p, bool strict = true) {
+    int a = 1, b = l.size() - 1, r = !strict;
```
simonlindholm: tabs vs spaces

fuzz-tests/geometry/PointInsideHull.cpp
Chillee: Leftover from debugging some bugs with the new impl.

PR #101, opened 2019-05-07
Moved minimum vertex cover next to bipartite matching

Chillee:
2 reasons:
1. MinimumVertexCover depends on the bipartite matching algorithms.
1. I think it makes more sense to put the bipartite matching stuff together, before the general graph matching.


simonlindholm:
Agreed.

PR #102, opened 2019-05-07
Updated Line Hull Intersection

Chillee:
I thought that Line Hull intersection took up a lot of space, especially considering that it's not too important.

This version has the following advantages:
1. 29 lines shorter
1. Doesn't require preprocessing of the convex hull.

Currently, there's some hijinks with the API. The original API is somewhat convoluted imo, perhaps we can do better than that? One option is to just return the points directly.

The tests are currently enabled to ensure that the "common case" (ie: proper intersections) matches the previous implementation's behavior. A bit more work is needed to ensure that the new function is correct for corner cases as well (I think it is...).


simonlindholm:
Very nice. Scary to hear about the broken test... Code looks good modulo tabs/spaces and corner-cases/tests. Header needs to be updated as well.

> I thought that Line Hull intersection took up a lot of space, especially considering that it's not too important.

Yeah... I don't think we put it in for any concrete reason either, more "tinykactl had this, and it might be useful to have".

> The original API is somewhat convoluted imo, perhaps we can do better than that? One option is to just return the points directly.

Well, it's complicated and full of cases because geometry is... We could maybe join some of cases, but I think it's pretty easy to deal with as a user. Returning points directly would make it impossible to use it for e.g. dynamical half-plane intersection.

Chillee:
At the cost of some really ugly checking, the half-plane code now passes the fuzz testing with the same outputs as the original code.

I'll work next on cleaning up the checking.

PR #103, opened 2019-05-07
Segment improvements

Tyilo:
Using `NEUTRAL` instead of `LOW` makes the code more clear when using a different function.

### DIFF
--- a/content/data-structures/SegmentTree.h
+++ b/content/data-structures/SegmentTree.h
@@ -3,7 +3,7 @@
  * Date: 2017-10-31
  * License: CC0
  * Source: folklore
- * Description: Zero-indexed max-tree. Bounds are inclusive to the left and exclusive to the right. Can be changed by modifying T, LOW and f.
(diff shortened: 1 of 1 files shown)


simonlindholm:
I dunno, I think LOW is pretty clear already, while NEUTRAL is longer. `constexpr` is also 4 chars longer than `const`; does it have any advantage? Using LOW instead of 0 in the constructor seems like a reasonable change, though.

Tyilo:
`constexpr` also works for non-integral T (`std::pair`).

Maybe we could rename `NEUTRAL` to `NEU`? Or `ID` for identity element?

simonlindholm:
Oh, working for pair sounds useful. NEU/ID both sound okay to me.

Chillee:
I like the name `unit`. At least conceptually, if you imagine the elements you're storing as a group, that's what the element should be (or: applying the element with the given operation does nothing).

simonlindholm:
`unit` sounds fine to me (probably clearer than NEU, and lower-case wins in being easy to type). Merging with that, LMK if you disagree.

PR #104, opened 2019-05-07
Only hash supported languages (c++)

Tyilo:


simonlindholm:
Can you include Java as well? `cpp` works there too. (For Python you start hitting problems with `//` -- it would be possible to skip hashing only when that happens, or more generally when the hash script fails, but maybe that's just confusing.)

simonlindholm:
Yes. The C preprocessor is harmless to run on most files, but Java especially so since it uses the same syntax for comments as C++.

simonlindholm:
Thanks! Fix the conflict and I'll merge.

Tyilo:
Fixed

PR #109, opened 2019-05-10
Added really really fast maximum clique implementation

Chillee:
Unlike the current `MaximalClique` (which finds all maximal cliques), this only finds the maximum clique in a graph. However, it does it unbelievably fast.

For `N=155` and random graphs with `p=0.90` (the worst case from my testing), this implementation solves them in ~1 second each.

For `N=63`, it can do about 4000 instances in 1 second. Compare that with the current one, which can do about... 2.

See: https://ideone.com/RznH1J

It is unfortunately fairly long (~58 lines currently). xyz/ecnerwala have mentioned that this is useful for hacking/cheesing problems, so perhaps it might be worth including. I tried to golf it but I'd appreciate any suggestions on where I might look to shorten it even more.

I chose this one over the MIT implementation because it appears to be significantly faster and can be extended past 63 nodes easily.


content/graph/MaximumClique.h
```diff
@@ -0,0 +1,62 @@
...
+ */
+struct Maxclique {
+	double limit=0.025, pk=0;
+	struct Vertex { int i, d; Vertex(int i) : i(i), d(0) {} };
+	typedef vector<Vertex> vv;
+	vector<B> e;
+	vv V;
+	vector<vi> C;
+	vi qmax, q, S, So;
+	bool cut1(const int pi, const vi &A) {
```
simonlindholm: `int pi, vi& A`

content/graph/MaximumClique.h
```diff
@@ -0,0 +1,62 @@
...
+		trav(v, r) trav(j, r) v.d += e[v.i][j.i];
+		sort(all(r), [](auto a, auto b) { return a.d > b.d; });
+	}
+	void expand(vv &R, int lev = 1) {
+		S[lev] += S[lev - 1] - So[lev];
+		So[lev] = S[lev - 1];
+		while (sz(R)) {
+			if (sz(q) + R.back().d <= sz(qmax)) return;
+			q.push_back(R.back().i);
+			vv Rp;
```
simonlindholm: `R`/`Rp` and `S`/`So` is easy to typo, and the mixed casing makes it hard to type. Rename to some other random capital letter?

content/graph/MaximumClique.h
```diff
@@ -0,0 +1,62 @@
...
+			} else if (sz(q) > sz(qmax)) qmax = q;
+			q.pop_back(), R.pop_back();
+		}
+	}
+	vi maxClique() { init(V), expand(V); return qmax; }
+	Maxclique(vector<B> conn) :
+        e(conn), S(sz(e)), So(S), C(sz(e) + 1) {
+		rep(i,0,sz(e)) V.push_back(Vertex(i));
+	}
+};
```
simonlindholm: newline at end of file

content/graph/MaximumClique.h
```diff
@@ -0,0 +1,62 @@
...
+ * https://github.com/FTRobbin/Dreadnought-Standard-Code-Library/blob/11c861e55a73be9c32a799bfe4398e0c62c2da15/improve/MaximumClique.cpp
+ * Description: Finds the maximum clique of a graph. Can be used to find the
+ * maximum independent set by finding the clique of the complement graph.
+ * Time: Runs in about ~1s for n=155 and worst case random graphs (p=.95). Runs
+ * faster for sparse graphs.
+ * Status: fuzz-tested
+ */
+struct Maxclique {
+	double limit=0.025, pk=0;
+	struct Vertex { int i, d; Vertex(int i) : i(i), d(0) {} };
```
simonlindholm: declare this as `struct Vertex { int i, d = 0; }` and construct instances using `{i}`

fuzz-tests/graph/MaximumClique.cpp
```diff
@@ -0,0 +1,41 @@
...
+	~timeit() {
+		auto end = chrono::high_resolution_clock::now();
+		auto duration = chrono::duration_cast<chrono::milliseconds>(end - begin).count();
+		cerr << duration << "ms elapsed [" << label << "]" << endl;
+	}
+};
+
+
+int main() {
+    rep(it, 0, 10000) {
```
simonlindholm: 10000 is too little

fuzz-tests/graph/MaximumClique.cpp
```diff
@@ -0,0 +1,41 @@
...
+        vector<B> ed(32);
```
simonlindholm: make `N` random between some limits

fuzz-tests/graph/MaximumClique.cpp
```diff
@@ -0,0 +1,41 @@
...
+        rep(i, 0, 32) rep(j, 0, i) {
+            ed[i][j] = (rand() % 100) < 90;
```
simonlindholm: and make the density parameter random as well

content/graph/MaximumClique.h
```diff
@@ -0,0 +1,62 @@
+/**
+ * Author: chilli, Haobin Ni
+ * Date: 2019-05-10
+ * License: CC0
```
simonlindholm: is CC0 true? if not, just remove this line

content/graph/MaximumClique.h
```diff
@@ -0,0 +1,62 @@
...
+ * Source: https://en.wikipedia.org/wiki/MaxCliqueDyn_maximum_clique_algorithm,
+ * https://github.com/FTRobbin/Dreadnought-Standard-Code-Library/blob/11c861e55a73be9c32a799bfe4398e0c62c2da15/improve/MaximumClique.cpp
+ * Description: Finds the maximum clique of a graph. Can be used to find the
+ * maximum independent set by finding the clique of the complement graph.
```
simonlindholm: the -> a

content/graph/MaximumClique.h
```diff
@@ -0,0 +1,62 @@
...
+ * Time: Runs in about ~1s for n=155 and worst case random graphs (p=.95). Runs
+ * faster for sparse graphs.
+ * Status: fuzz-tested
+ */
+struct Maxclique {
+	double limit=0.025, pk=0;
+	struct Vertex { int i, d; Vertex(int i) : i(i), d(0) {} };
+	typedef vector<Vertex> vv;
+	vector<B> e;
```
simonlindholm: hm, missing a definition of `B`. Use `/// include-line` if needed for the fuzz-test

content/graph/MaximumClique.h
Chillee: Yeah it's also from the SJTU notebook.

content/graph/MaximumClique.h
Chillee: Perhaps it should just be a `vector<bool>`?

simonlindholm: bitset is fewer indirections, so I'd stick with that. Hm, and defining it within the struct should avoid the need for  `/// include-line`

content/graph/MaximalCliques.h
```diff
@@ -13,7 +13,7 @@
  */
 #pragma once
 
-typedef bitset<128> B;
+typedef bitset<63> B;
```
simonlindholm: why 63?

fuzz-tests/graph/MaximumClique.cpp
```diff
@@ -0,0 +1,55 @@
...
+void cliques(vector<B>& eds, F f, B P = ~B(), B X={}, B R={}) {
+	if (!P.any()) { if (!X.any()) f(R); return; }
+	auto q = (P | X)._Find_first();
+	auto cands = P & ~eds[q];
+	rep(i,0,sz(eds)) if (cands[i]) {
+		R[i] = 1;
+		cliques(eds, f, P & eds[i], X & eds[i], R);
+		R[i] = P[i] = 0; X[i] = 1;
+	}
+}
```
simonlindholm:
```
namespace other {
#include "..."
}
```
is another option

fuzz-tests/graph/MaximumClique.cpp
```diff
@@ -0,0 +1,55 @@
...
+	~timeit() {
+		auto end = chrono::high_resolution_clock::now();
+		auto duration = chrono::duration_cast<chrono::milliseconds>(end - begin).count();
+		cerr << duration << "ms elapsed [" << label << "]" << endl;
+	}
+};
+
+
+int main() {
+    int n =rand()%32;
```
simonlindholm: move this within the loop

content/graph/MaximumClique.h
Chillee: I mean, `S` and `So` stand for `S` and `S old`, so I think they should at least start with the same name.

content/graph/MaximalCliques.h
Chillee: Misremembered what the old value was.

content/graph/MaximumClique.h
simonlindholm: `old`? or some two consecutive letters?

PR #110, opened 2019-05-11
Added directedMST/Minimum arborescence code

Chillee:
I ended up choosing [Suprdewd's implementation](https://github.com/SuprDewd/CompetitiveProgramming/blob/master/code/graph/arborescence.cpp) largely because I wanted to be able to get the edges that were a part of the MST as well. It's not clear to me that it's important, but it might show up, and the problem that originally motivated me to find an implementation [required it](https://codeforces.com/contest/240/problem/E). MIT's implementation seems to be a fairly popular one, but unluckily, it doesn't seem that I'm unable to modify it to return edges without a significant increase in complexity (and memory usage I believe).

I'm very open to opinions about the usability/API of this. Currently it returns a pair of (min cost, edge indices), where edges must be added using an `addEdge` function.

I'll also probably add a fuzz test comparing against the MIT style implementation.

Honestly, I find the code pretty ugly - but it is what it is. There's just a lot of steps involved.


Chillee:
If we don't care about getting the edges back, the code can be simplified significantly.

Chillee:
Hmm....

Troublingly, in addition to taking `VE` time, I think it also takes `VE` memory (in order to recover the actual edges).

I'm starting to think that recovering the edges isn't worth it, sadly, unless there's another way to do it that's not the way I'm currently doing it.

Chillee:
Also, more precisely on recovering the edges, this is a paper that explains how to do it: https://cw.fel.cvut.cz/old/_media/courses/a4m33pal/cviceni/carmerini-a-note-on-finding-optimum-branchings.pdf

Chillee:
I decided to just remove the `VE` implementation altogether - the one currently in the repo isn't great. It takes `VE` memory and is nearly as long as the `E log V` one.

PR #118, opened 2019-05-15
Added a test runner

Chillee:
2 changes (other than minor test output changes):
1. Fixed segtree tests that were broken after the recent update.
2. Added a test runner that runs all the tests (ie: `.cpp` files in `fuzz-tests). 

Finally outputs this:
image

### DIFF
--- a/runtests.sh
+++ b/runtests.sh
@@ -1,8 +1,28 @@
-tests="$(find fuzz-tests -name '*.cpp')"
+DIR=${1:-fuzz-tests}
+tests="$(find $DIR -name '*.cpp')"
+declare -i pass=0
+declare -i fail=0
+failTests=""
+echo -e $failTests
 for test in $tests; do
-    basename $test
-    g++ -std=c++14 $test
+    echo "$(basename $test): "
+    g++ -std=c++14 -O2 $test
     ./a.out
-    echo $?
+    retCode=$?
+    if (($retCode != 0)); then
+        echo $retCode
+        fail+=1
+        failTests="$failTests$test\n"
+        echo -e $failTests
+    else
+        pass+=1
+    fi
     rm a.out
+    echo ""
 done
+echo "$pass/$(($pass+$fail)) tests passed"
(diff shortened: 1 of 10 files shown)


runtests.sh
```diff
@@ -0,0 +1,28 @@
...
+for test in $tests; do
+    echo "$(basename $test): "
+    g++ -std=c++14 -O2 $test
+    ./a.out
+    retCode=$?
+    if (($retCode != 0)); then
+        echo $retCode
+        fail+=1
+        failTests="$failTests$test\n"
+        echo -e $failTests
```
simonlindholm: this line probably shouldn't be here

runtests.sh
```diff
@@ -0,0 +1,28 @@
+DIR=${1:-fuzz-tests}
+tests="$(find $DIR -name '*.cpp')"
+declare -i pass=0
+declare -i fail=0
+failTests=""
+echo -e $failTests
+for test in $tests; do
+    echo "$(basename $test): "
+    g++ -std=c++14 -O2 $test
+    ./a.out
```
simonlindholm: `g++ -std=c++14 -O2 $test && ./a.out` to catch compile errors as well?

runtests.sh
```diff
@@ -0,0 +1,28 @@
...
+    retCode=$?
+    if (($retCode != 0)); then
+        echo $retCode
+        fail+=1
+        failTests="$failTests$test\n"
+        echo -e $failTests
+    else
+        pass+=1
+    fi
+    rm a.out
```
simonlindholm: `rm -f` for compile errors' sake

runtests.sh
```diff
@@ -0,0 +1,28 @@
...
+    echo ""
```
simonlindholm: can remove the `""`

simonlindholm:
Hm, can you add a `#!/usr/bin/env bash` line at the top as well. I'll merge with that.

PR #119, opened 2019-05-18
Added description of MaximumIndependentSet.h and some other minor description tweaks.

Chillee:
Considering that maxclique is often used for this problem, it's probably worth adding it in the description. On the other hand, it's simple/obvious enough that I'd be fine with simply appending it to the end of the clique code.

Also made some other miscellaneous changes to be more clear and make the API more obvious.


simonlindholm:
Sure, seems worth it just for the note about bipartite graphs.

Issue #117, opened 2019-05-14
Augmenting LinkCutTree doesn't seem to work


Tyilo:
I have tried to augment LinkCutTree with the subtree size, however I can't get it to work.

I have added `int sz = 1;` as a field to `Node` and modified `Node::fix()` to contain:
```
void fix() {
	sz = 1;
	if (c[0]) {
		c[0]->p = this;
		sz += c[0]->sz;
	}
	if (c[1]) {
		c[1]->p = this;
		sz += c[1]->sz;
	}
	// (+ update sum of subtree elements etc. if wanted)
}
```

My test code is
```
LinkCut T(4);
T.link(1, 0);
T.link(2, 1);
T.link(3, 0);

rep(i, 0, 4) {
	cout << i << " " << T.node[i].sz << "\n";
}
```

I would expect the output to be:
```
0 4
1 2
2 1
3 1
```

But instead I get:
```
0 2
1 1
2 1
3 1
```

The full code is available at https://gist.github.com/Tyilo/ebbbd06dcb02665c078ec3dac387c09a

simonlindholm:
Ugh. I don't understand the link-cut tree well enough to say what's going wrong. We should replace it by a better version... (#63)

Chillee:
In general, I think for data structures that can be augmented, like segtrees or (as we recently learned) Link Cut Trees, we should try to separate the code that's meant to be augmented from the rest of the code.

Altynai:
@Tyilo you might need to maintain the size of virtual children as well, see Benq's implementation: [LCT.h#L27](https://github.com/bqi343/USACO/blob/master/Implementations/content/graphs%20(12)/Advanced/LCT.h#L27=), example problem: https://codeforces.com/contest/1681/problem/F

PR #124, opened 2019-07-02
Link Cut Tree

PotatoHashing:
This PR is about updating the Link Cut Tree. Most of it is from MIT's LCT. This new version easily supports dynamic links and cuts, as well as queries and updates on paths (like HLD). The code has been separated so that the functions in the Node struct will change the structure of the LCT (access, splay, rotate, etc) and the functions under LinkCut will call those under Node to perform queries and updates about the tree. Alternatively, the functions under LinkCut are meant to be called by a user directly (e.g. in main) while the functions in Node aren't.

The key points are:
1. Access(u) before updating/querying a node u.
2. Make root(u) and access(v) before updating/querying a path.

The current code shows an example on how to update a point on a path and query for max on a path. (It basically solves http://acm.timus.ru/problem.aspx?num=1553).

Some current possible issues.
1. xiao can probably be renamed. xiao(x) basically does "go up the current splay tree that x is in, and make all nodes on that path have the same path parent as the root of that splay tree".
2. Mixture of "." and "->" in LinkCut due to sometimes using Node and other time using Node*. Might be helpful to standardize.
3. /// start-hash and /// end-hash were left over from MIT.
4. Cut and link use makeRoot to simplify their implementations. This means that if a represented tree has fixed root, it would be necessary to call an additional makeRoot to reroot the tree correctly after each link/cut. This could also be an issue for problems of HLD on edges where the value of a node stores the value of the edge from that node to its parent. (Although in the HLD case, we can just create N-1 additional nodes, one for each edge). I've tried to see what changes need to be made to represent a fixed rooted LCT, but it seems non trivial at the moment.

Future work.
There's this blog on applications on LCT. It might be useful to include parts of that. Unfortunately, I can't read Chinese, so I'm leaving this incomplete for now.
https://www.cnblogs.com/flashhu/p/8324551.html

I am open to questions and suggestions.

### DIFF
--- a/content/graph/LCT.h
+++ b/content/graph/LCT.h
@@ -70,35 +70,38 @@ struct Node {
 			z->pull();
 		}
 		splay();
-		flip = 1;
-		push();
 		return this;
 	} /// end-hash
 };
 struct LinkCut {
 	vector<Node> node;
 	LinkCut(int N): node(N) {}
+	Node* makeRoot(int u) {
+		node[u].access()->flip ^= 1;
+		return &node[u];
+	}
+	Node* findRoot(int u) {
+		Node *x = node[u].access();
+		while(x->c[0]) x = x->c[0];
+		return x;
+	}
 	bool cut(int u, int v) { /// start-hash
 		Node *x = &node[u], *y = &node[v];
+		makeRoot(v);
 		x->access();
-		y->access();
-		if (y->c[1] != x || x->c[0] != 0)
+		if (x->c[0] != y || y->c[1] != 0)
 			return false;
-		y->c[1] = 0;
-		x->p = x->pp = 0;
-		y->pull();
+		x->c[0] = y->p = y->pp = 0;
+		x->pull();
 		return true;
 	} /// end-hash
 	bool isConnected(int u, int v) {
-		Node *x = &node[u], *y = &node[v];
-		x->access();
-		y->access();
-		return x == y || x->p != 0;
+		return findRoot(u) == findRoot(v);
 	}
 	bool link(int u, int v) {
-		Node *x = &node[u], *y = &node[v];
+		Node *y = &node[v];
 		if (isConnected(u, v)) return false;
-		x->access()->pp = y;
+		makeRoot(u)->pp = y;
 		return true;
 	}
 	// Add c to node u.
@@ -107,7 +110,7 @@ struct LinkCut {
 	}
 	// Find max on the path from u to v.
 	ll query(int u, int v) {
-		node[v].access();
+		makeRoot(v);
 		return node[u].access()->cval;
 	}
 };
--- a/content/graph/LCT.h
+++ b/content/graph/LCT.h
@@ -1,6 +1,11 @@
 /**
  * Author: 
- * Description: link-cut Tree.
+ * Description: link-cut Tree. Supports BBST-like 
+ * augmentations. (Can be used in place of HLD).
+ * Current implementation supports update value at a node,
+ * and query max on a path.
+ * Tested on: http://acm.timus.ru/problem.aspx?num=1553
+ * Status: Passes existing fuzz tests (with function names modified).
  */
 struct Node {
 	bool flip = 0;
@@ -29,12 +34,12 @@ struct Node {
 		Node *y = p, *z = y->p;
 		if (z) z->c[z->c[1] == y] = this;
 		p = z;
-		if (y) y->c[t] = c[!t];
+		y->c[t] = c[!t];
 		if (c[!t]) c[!t]->p = y;
 		c[!t] = y;
 		y->p = this;
 		if (z) z->pull();
-		if (y) y->pull();
+		y->pull();
 	} /// end-hash
 	void xiao() {
 		if (p) p->xiao(), pp = p->pp;
--- a/content/graph/LCT.h
+++ b/content/graph/LCT.h
@@ -16,6 +16,7 @@ struct Node {
 		pp = p = c[0] = c[1] = 0; 
 		val = cval = 0;
 	}
+	// lazy propogation
 	void push() {
 		if (flip) {
 			if(c[0]) c[0]->flip ^= 1;
@@ -24,8 +25,9 @@ struct Node {
 		}
 		// add stuff
 	}
+	// combine values from children with self.
 	void pull() {
-        cval = val;
+		cval = val;
 		if(c[0]) c[0]->push(), cval = max(cval, c[0]->cval); 
 		if(c[1]) c[1]->push(), cval = max(cval, c[1]->cval);
 		// add stuff
@@ -96,13 +98,14 @@ struct LinkCut {
 	bool link(int u, int v) {
 		Node *x = &node[u], *y = &node[v];
 		if (isConnected(u, v)) return false;
-		x->access();
-		x->pp = y;
+		x->access()->pp = y;
 		return true;
 	}
+	// Add c to node u.
 	void update(int u, int c) {
 		node[u].access()->val+=c;
 	}
+	// Find max on the path from u to v.
 	ll query(int u, int v) {
 		node[v].access();
 		return node[u].access()->cval;
--- a/content/graph/LCT.h
+++ b/content/graph/LCT.h
@@ -1,0 +1,111 @@
+/**
+ * Author: 
+ * Description: link-cut Tree.
+ */
+struct LinkCut {
+  struct Node {
+    bool flip = 0;
+    Node *pp, *p, *c[2];
+    // add stuff
+  };
+
+  vector<Node> node;
+  vector<Node*> f;
+  Node *null;
+	LinkCut(int N) : node(N), f(N), null(new Node) {
+    null -> c[0] = null -> c[1] = null -> p = null -> pp = null;
+    rep(i, 0, N) {
+      f[i] = &node[i];
+      f[i] -> c[0] = f[i] -> c[1] = f[i] -> p = f[i] -> pp = null;
+    }
+  }
+
+  void push(Node *x) {
+    if (x -> flip) {
+      x -> c[0] -> flip ^= 1;
+      x -> c[1] -> flip ^= 1;
+      swap(x -> c[0], x -> c[1]);
+      x -> flip = false;
+    }
+    // add stuff
+  }
+
+  void pull(Node *x) {
+    push(x -> c[0]);
+    push(x -> c[1]);
+    // add stuff
+  }
+
+  void rot(Node *x, bool t) { /// start-hash
+    Node *y = x -> p, *z = y -> p;
+    if (z != null)  z -> c[z -> c[1] == y] = x;
+    x -> p = z;
+    y -> c[t] = x -> c[!t];
+    x -> c[!t] -> p = y;
+    x -> c[!t] = y;
+    y -> p = x;
+    pull(z);
+    pull(y);
+  } /// end-hash
+
+  void xiao(Node *x) {
+    if (x -> p != null)  xiao(x -> p), x -> pp = x -> p -> pp;
+    push(x);
+  }
+
+  void splay(Node *x) { /// start-hash
+    xiao(x);
+    Node *y, *z;
+    while (x -> p != null) {
+      y = x -> p; z = y -> p;
+      bool t1 = (y -> c[1] == x), t2 = (z -> c[1] == y);
+      if (z != null) {
+        if (t1 == t2) rot(y, t2), rot(x, t1);
+        else  rot(x, t1), rot(x, t2);
+      } else rot(x, t1);
+    }
+    pull(x);
+  } /// end-hash
+
+  Node* access(Node *x) {/// start-hash
+    for (Node *y = null, *z = x; z != null; y = z, z = z -> pp) {
+        splay(z);
+        z -> c[1] -> pp = z;
+        z -> c[1] -> p = null;
+        z -> c[1] = y;
+        y -> p = z;
+        push(z);
+    }
+    splay(x);
+    x -> flip = true;
+		push(x);
+    return x;
+  } /// end-hash
+
+  bool cut(int u, int v) { /// start-hash
+    access(f[u]);
+    access(f[v]);
+    if (f[v] -> c[1] != f[u] || f[u] -> c[0] != null)
+      return false;
+    f[v] -> c[1] = null;
+    f[u] -> p = f[u] -> pp = null;
+    pull(f[u]);
+    pull(f[v]);
+    return true;
+  } /// end-hash
+
+  bool isConnected(int u, int v) {
+    access(f[u]);
+    access(f[v]);
+    return f[u] == f[v] || f[u] -> p != null;
+  }
+
+  bool link(int u, int v) {
+    if (isConnected(u, v))
+      return false;
+    access(f[u]);
+    access(f[v]);
+    f[u] -> pp = f[v];
+    return true;
+  }
+};
--- a/content/graph/LCT.h
+++ b/content/graph/LCT.h
@@ -2,110 +2,94 @@
  * Author: 
  * Description: link-cut Tree.
  */
+struct Node {
+	bool flip = 0;
+	Node *pp, *p, *c[2];
+	// add stuff
+	Node() { pp = p = c[0] = c[1] = 0; }
+	void push() {
+		if (flip) {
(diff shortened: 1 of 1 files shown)


content/graph/LCT.h
```diff
@@ -0,0 +1,116 @@
+/**
+ * Author: 
+ * Description: link-cut Tree. Supports BBST-like
```
simonlindholm: BBST -> BST?

content/graph/LCT.h
```diff
@@ -0,0 +1,116 @@
+/**
+ * Author: 
+ * Description: link-cut Tree. Supports BBST-like 
+ * augmentations. (Can be used in place of HLD).
+ * Current implementation supports update value at a node,
+ * and query max on a path.
+ * Tested on: http://acm.timus.ru/problem.aspx?num=1553
+ * Status: Passes existing fuzz tests (with function names modified).
```
simonlindholm: We'll want to remove the old link-cut tree, so feel free to do that and modify the fuzz-test. Will also need to update the test to verify that max works (with max replaced by some non-commutative, non-associative function)

PotatoHashing:
Also, the current LCT doesn't fix #117. It would need the changes from https://codeforces.com/blog/entry/67637 or the Chinese blog.

content/graph/LCT.h
```diff
@@ -0,0 +1,116 @@
...
+struct Node {
+	bool flip = 0;
+	Node *pp, *p, *c[2];
+	// add stuff
+	int val, cval;
+	Node() { 
+		pp = p = c[0] = c[1] = 0; 
+		val = cval = 0;
+	}
+	// lazy propogation
```
simonlindholm: remove this comment

content/graph/LCT.h
```diff
@@ -0,0 +1,116 @@
...
+	void push() {
+		if (flip) {
+			if(c[0]) c[0]->flip ^= 1;
+			if(c[1]) c[1]->flip ^= 1;
+			swap(c[0], c[1]); flip = 0;
+		}
+		// add stuff
```
simonlindholm: remove this comment as well -- I think we can instead write in the description how to augment it ("extend push/pull").

We'll also want a more thorough explanation and/or example -- it's not clear to me in which ways the LCT is similar to a segment tree. E.g. how do I do lazy updates on entire paths in the tree? Entire subtrees? Queries on subtrees? Can I use the flip bit myself for anything? How are things ordered? It's not reasonable to answer all of these questions (we should have a link in the description!), but some more info would be useful.

content/graph/LCT.h
```diff
@@ -0,0 +1,116 @@
...
+	}
+	// combine values from children with self.
+	void pull() {
+		cval = val;
+		if(c[0]) c[0]->push(), cval = max(cval, c[0]->cval); 
+		if(c[1]) c[1]->push(), cval = max(cval, c[1]->cval);
+		// add stuff
```
simonlindholm: Does this work with non-commutative and non-associative operations?

content/graph/LCT.h
```diff
@@ -0,0 +1,116 @@
...
+	int val, cval;
+	Node() { 
+		pp = p = c[0] = c[1] = 0; 
+		val = cval = 0;
+	}
+	// lazy propogation
+	void push() {
+		if (flip) {
+			if(c[0]) c[0]->flip ^= 1;
+			if(c[1]) c[1]->flip ^= 1;
```
simonlindholm: Probably worth doing `rep(i,0,2)`

content/graph/LCT.h
```diff
@@ -0,0 +1,116 @@
...
+ * Description: link-cut Tree. Supports BBST-like 
+ * augmentations. (Can be used in place of HLD).
+ * Current implementation supports update value at a node,
+ * and query max on a path.
+ * Tested on: http://acm.timus.ru/problem.aspx?num=1553
+ * Status: Passes existing fuzz tests (with function names modified).
+ */
+struct Node {
+	bool flip = 0;
+	Node *pp, *p, *c[2];
```
simonlindholm: is it possible to add a one-line comment explaining what p and pp are? or are they standard names?

content/graph/LCT.h
```diff
@@ -0,0 +1,116 @@
...
+	// add stuff
+	int val, cval;
```
simonlindholm: `int val = 0, cval = 0;` etc. to get rid of the constructor

content/graph/LCT.h
```diff
@@ -0,0 +1,116 @@
...
+	void rot(bool t) { /// start-hash
+		Node *y = p, *z = y->p;
+		if (z) z->c[z->c[1] == y] = this;
+		p = z;
+		y->c[t] = c[!t];
+		if (c[!t]) c[!t]->p = y;
+		c[!t] = y;
+		y->p = this;
+		if (z) z->pull();
+		y->pull();
```
simonlindholm: this is an unreadable mess, it wouldn't hurt to golf it a bit and make it even more unreadable... some ideas:
- !t is used more than t, maybe invert its meaning
- `, *&w = y->c[!t]` at the top and use `w` instead of `c[!t]`
- putting things on the same line is possibly worth it, if done tastefully... e.g. maybe `y->p = this` and `p = z` could go together?
- some sort of "fix this node up" function which is called on some subset of c[!t]/this/y/z

Also, I'm a bit surprised the pull for z happens before that of y, I would have expected the other way around.

content/graph/LCT.h
```diff
@@ -0,0 +1,116 @@
...
+		xiao();
+		Node *y, *z;
+		while (p) {
+			y = p; z = y->p;
+			bool t1 = (y->c[1] == this);
+			if (z) {
+				bool t2 = (z->c[1] == y);
+				if (t1 == t2) y->rot(t2), rot(t1);
+				else rot(t1), rot(t2);
+			} else rot(t1);
```
simonlindholm: how about:
```
bool t2 = z && (z->c[1] == y) == t1;
if (t2) y->rot(t1);
rot(t1);
if (z && !t2) rot(!t1);
```
? arguably uglier, but saves a line

content/graph/LCT.h
```diff
@@ -0,0 +1,116 @@
...
+	} /// end-hash
+	void xiao() {
+		if (p) p->xiao(), pp = p->pp;
+		push();
+	}
+	void splay() { /// start-hash
+		xiao();
+		Node *y, *z;
+		while (p) {
+			y = p; z = y->p;
```
simonlindholm: `Node *y = .., *z = ..;` to win a line

content/graph/LCT.h
```diff
@@ -0,0 +1,116 @@
...
+		// add stuff
+	}
+	// combine values from children with self.
+	void pull() {
+		cval = val;
+		if(c[0]) c[0]->push(), cval = max(cval, c[0]->cval); 
+		if(c[1]) c[1]->push(), cval = max(cval, c[1]->cval);
+		// add stuff
+	}
+	void rot(bool t) { /// start-hash
```
simonlindholm: remove the `start-hash`/`end-hash` comments

content/graph/LCT.h
```diff
@@ -0,0 +1,116 @@
...
+			z->c[1] = y;
+			if (y) y->p = z;
+			z->pull();
+		}
+		splay();
+		return this;
+	} /// end-hash
+};
+struct LinkCut {
+	vector<Node> node;
```
simonlindholm: `nodes`

content/graph/LCT.h
```diff
@@ -0,0 +1,116 @@
...
+	LinkCut(int N): node(N) {}
+	Node* makeRoot(int u) {
```
simonlindholm: might as well make this void, I think

content/graph/LCT.h
```diff
@@ -0,0 +1,116 @@
...
+	Node* findRoot(int u) {
+		Node *x = node[u].access();
+		while(x->c[0]) x = x->c[0];
+		return x;
+	}
+	bool cut(int u, int v) { /// start-hash
+		Node *x = &node[u], *y = &node[v];
+		makeRoot(v);
+		x->access();
+		if (x->c[0] != y || y->c[1] != 0)
```
simonlindholm: remove `!= 0`

content/graph/LCT.h
```diff
@@ -0,0 +1,116 @@
...
+			return false;
+		x->c[0] = y->p = y->pp = 0;
+		x->pull();
+		return true;
+	} /// end-hash
+	bool isConnected(int u, int v) {
+		return findRoot(u) == findRoot(v);
+	}
```
simonlindholm: put this function on one line

content/graph/LCT.h
```diff
@@ -0,0 +1,116 @@
...
+		if (isConnected(u, v)) return false;
+		makeRoot(u)->pp = y;
+		return true;
+	}
+	// Add c to node u.
+	void update(int u, int c) {
+		node[u].access()->val+=c;
+	}
+	// Find max on the path from u to v.
+	ll query(int u, int v) {
```
simonlindholm: `int`, and maybe move the comment onto the same line as the function header

content/graph/LCT.h
```diff
@@ -0,0 +1,116 @@
...
+	}
+	bool link(int u, int v) {
+		Node *y = &node[v];
+		if (isConnected(u, v)) return false;
+		makeRoot(u)->pp = y;
+		return true;
+	}
+	// Add c to node u.
+	void update(int u, int c) {
+		node[u].access()->val+=c;
```
simonlindholm: space around `+=`, and same comment about the comment as below (or maybe remove it altogether)

content/graph/LCT.h
```diff
@@ -0,0 +1,116 @@
...
+	}
+	// Find max on the path from u to v.
+	ll query(int u, int v) {
+		makeRoot(v);
+		return node[u].access()->cval;
+	}
+};
```
simonlindholm: (super-minor nit, but you may want a newline at the end of file)

content/graph/LCT.h
```diff
@@ -0,0 +1,116 @@
...
+			if (y) y->p = z;
+			z->pull();
+		}
+		splay();
+		return this;
+	} /// end-hash
+};
+struct LinkCut {
+	vector<Node> node;
+	LinkCut(int N): node(N) {}
```
simonlindholm: space around `:`

content/graph/LCT.h
```diff
@@ -0,0 +1,116 @@
...
+		if (z) z->c[z->c[1] == y] = this;
+		p = z;
+		y->c[t] = c[!t];
+		if (c[!t]) c[!t]->p = y;
+		c[!t] = y;
+		y->p = this;
+		if (z) z->pull();
+		y->pull();
+	} /// end-hash
+	void xiao() {
```
simonlindholm: I agree that we should probably rename this

simonlindholm:
Neat. I think the main thing I would want out of this is a more detailed view of how it can be augmented. We should try not to bloat it too much with unnecessary cruft, but right now it's hard for me to understand how to augment it with non-trivial custom stuff.

I wouldn't worry too much about point 4 if it's hard to figure out, keeping it simple is good. And I don't think mixing . and -> is a problem.

PotatoHashing:
I forgot to push my most recent changes before making the PR. It's pushed now.

content/graph/LCT.h
PotatoHashing: > , *&w = y->c[!t] at the top and use w instead of c[!t]

I might be misunderstanding, but why *&w instead of *w?
Also y->c[!t] is only used once, while this->c[!t] is used multiple times, so *&w = c[!t] might be better.

> some sort of "fix this node up" function which is called on some subset of c[!t]/this/y/z

Like the fix function in the previous lct? (Set a node's children, and then fix will set those children's parents, but without the "(+ update sum of subtree elements etc. if wanted)" part, since that is what pull does).

Pull for z was unnecessary. (A node is only moved up one level for each call of rot, and then pulls up itself at the end of splay.) Thanks for noticing.

Everything else seems good.

content/graph/LCT.h
PotatoHashing: Updated the fuzz-test to verify max for now. Will update it later for any new function that we decide on...

content/graph/LCT.h
PotatoHashing: p = parent, pp = path-parent (and c = children).
p and c are straight forward. path-parent is from the Wikipedia article on LCTs and also this set of lecture notes https://courses.csail.mit.edu/6.851/spring12/scribe/L19.pdf (which I think the Wikipedia page is based on). Besides writing out the acronyms, do you think it's a good idea to explain what a path-parent is? I think that if you understand how LCTs work (by decomposing a tree into preferred paths), then knowing that pp stands for path-pointer should be enough information.

content/graph/LCT.h
PotatoHashing: I think that often when calling makeRoot on a node, you have to use (update or get information from) that node soon after, so it's beneficial to leave as is. But adding a ```Node *x = &node(u);``` after ```makeRoot(u);``` wouldn't be too much of a hassle either.

content/graph/LCT.h
PotatoHashing: Changed, but goes slightly over the line limit now.

content/graph/LCT.h
PotatoHashing: "propogatePathParent" is a little much though... Will keep thinking about possible alternatives.

content/graph/LCT.h
simonlindholm: So something like `// p ~ parent in segtree, pp = path parent`? (Now, I'm also not sure whether that's helpful information when augmenting... but I could imagine such a comment being helpful for understanding)

content/graph/LCT.h
simonlindholm: could certainly shorten that to `propPP` if that's what we want to get across. But a nonsensical name would be fine as well

content/graph/LCT.h
simonlindholm: At the same time it adds some confusion, because it makes you think that the return value is something magical that you need to use while in actuality it's just the same as the input. (Well, I suppose not a lot of confusion, since you're typing the whole thing in and see the `return this;`.) Also it seems to make the code slightly longer right now. It's not a big deal, but I think we'd be slightly better off without the "return this".

content/graph/LCT.h
simonlindholm: oh, I confused `y->c[!t]` and `c[!t]`. `c[!t]` is assigned to, hence the `&`.

> Like the fix function in the previous lct?

Yeah. I don't know much about how the previous lct works, but it seems like a method along those lines *may* be able to cut away a few lines? Again, I'm just throwing ideas out, I don't know whether it actually does save anything.

content/graph/LCT.h
simonlindholm: Hrm, that's no good then... Either revert or change the function names

content/graph/LCT.h
PotatoHashing: What about integrating the functionality of makeRoot into access by adding a boolean parameter and then getting rid of makeRoot entirely? Something like
```
Node* access(bool r) {
...
    splay();
    if (r) flip ^= 1, push();
    return this;
}
```

simonlindholm: Sounds reasonable to me.

content/graph/LCT.h
PotatoHashing: Doesn't look like adding fix will shorten it since `w->p = y;` is already as short as `w->fix();`. And then sometimes we'll have to set a parent directly anyway. (when `z = 0` and we need to set `this->p = z` then we need to set `p = z` without calling `z->fix()` anyway). I think it's ok to leave as it is. 
The other changes have been put in.

content/graph/LCT.h
PotatoHashing: reverted

content/graph/LCT.h
PotatoHashing: We've talked about this elsewhere, but it should work for non-commutative, associative functions (like a segment tree).

Chillee:
@PotatoHashing would you still like to finish this PR? I want to get a LCT implementation - perhaps ideally this one. Alternately, I could take over for this PR if you're busy.

PotatoHashing:
We discussed this on Discord but I should probably repeat it here. Feel free to take this over @Chillee (or anyone else interested).

PR #125, opened 2019-07-19
Changes to Aho-Corasick

simonlindholm:
I thought I needed an Aho-Corasick with map instead of array to support large alphabets, and started working on some changes to that effect (mostly defaulting to 0 instead of -1 for non-existent trie entries). Turns out I don't know how to make the tree post-processing step work without filling in all entries; the best idea I have so far is to give up on maps and instead split symbols into chunks, with markers for initial chunks.

Some of the changes here are probably useful, but swapping back to 0/-1 may be good (1 is an unnatural starting index, and memset is faster than normal array initialization on GCC). Either way, I'm dropping this for now and making a PR as a reminder for later.

### DIFF
--- a/content/strings/AhoCorasick.h
+++ b/content/strings/AhoCorasick.h
@@ -3,68 +3,67 @@
  * Date: 2015-02-18
  * License: CC0
  * Source: marian's (TC) code
- * Description: Aho-Corasick tree is used for multiple pattern matching.
- * Initialize the tree with create(patterns). find(word) returns for each position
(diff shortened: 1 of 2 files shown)


simonlindholm:
Pushed the description changes and the stress-test in 9c69215eff107c416cfd325fea137d25b2f1150d and 4a147ce8e00009592a7a1adcf6ef9c10493a063b; not worth keeping this PR open for the rest.

PR #130, opened 2019-10-23
Added short explanation to linedistance

Chillee:
Took a little bit of thinking - probably worth adding a short part about this.


simonlindholm:
Agreed. (My intuition is that you can rotate the 3d case into 2d, and that doesn't affect cross product magnitudes.)

PR #131, opened 2019-10-23
Added time complexity for berlekamp massey

Chillee:
Don't see any reason to not include this.


simonlindholm:
The modpow. In practice the mod is constant, but I'm thinking maybe it's useful to see that there's a large constant factor (~50 mod multiplications per inversion)?

Chillee:
The `modpow` isn't in the inner loop, so I doubt it matters. 

It's `O(N^2 + N log MOD)`.

simonlindholm:
Oh, oops, totally misread then... Will merge.

PR #134, opened 2019-10-30
Added polygon area union

Chillee:
Fuzz tested.

There exists a `n log n + k` algorithm, where k is the number of intersections, but I think it's significantly more complicated.


content/geometry/PolygonUnionArea.h
```diff
@@ -0,0 +1,49 @@
+/**
+ * Author: black_horse2014, chilli
+ * Date: 2019-10-29
+ * License: CC0
+ * Source:
+ * Description: Calculates the area of the union of $n$ polygons (not necessarily convex).
```
simonlindholm: Does it require anything of the polygons' orientations?

content/geometry/PolygonUnionArea.h
```diff
@@ -0,0 +1,49 @@
...
+		rep(j,0,sz(poly)) {
+			if (i == j) continue;
+			rep(u,0,sz(poly[j])) {
+				P C = poly[j][u], D = poly[j][(u + 1) % sz(poly[j])];
+				int sc = sideOf(A, B, C), sd = sideOf(A, B, D);
+				if (sc != sd) {
+					double sa = C.cross(D, A), sb = C.cross(D, B);
+					if (min(sc, sd) < 0)
+						segs.emplace_back(sa / (sa - sb), sgn(sc - sd));
+				} else if (!sc && !sd && j<i && sgn((B-A).dot(D-C))>0){
```
simonlindholm: There's comparisons being done on doubles everywhere here, so I guess a requirement is that the coordinates all be integers < ~2^25?

content/geometry/PolygonUnionArea.h
```diff
@@ -0,0 +1,49 @@
...
+double rat2(P a, P b) { return sgn(b.x) ? a.x/b.x : a.y/b.y; }
+double polyUnion(vector<vector<P>> &poly) {
+	double ret = 0;
+	rep(i,0,sz(poly)) rep(v,0,sz(poly[i])) {
+		P A = poly[i][v], B = poly[i][(v + 1) % sz(poly[i])];
+		vector<pair<double, int>> segs({{0, 0}, {1, 0}});
+		rep(j,0,sz(poly)) {
+			if (i == j) continue;
+			rep(u,0,sz(poly[j])) {
+				P C = poly[j][u], D = poly[j][(u + 1) % sz(poly[j])];
```
simonlindholm: There's a lot of `poly[i]` and `poly[j]` and `poly` here, which are long and easily confusable. How about `auto &A = poly[i], &B = poly[j]`? (... well, ok, with other names because you're already using A and B)

content/geometry/PolygonUnionArea.h
```diff
@@ -0,0 +1,49 @@
...
+
+typedef Point<double> P;
+double rat2(P a, P b) { return sgn(b.x) ? a.x/b.x : a.y/b.y; }
+double polyUnion(vector<vector<P>> &poly) {
+	double ret = 0;
+	rep(i,0,sz(poly)) rep(v,0,sz(poly[i])) {
+		P A = poly[i][v], B = poly[i][(v + 1) % sz(poly[i])];
+		vector<pair<double, int>> segs({{0, 0}, {1, 0}});
+		rep(j,0,sz(poly)) {
+			if (i == j) continue;
```
simonlindholm: can be a filtered rep, `rep(j,0,sz(poly)) if (i != j) {`

fuzz-tests/geometry/PolygonUnionArea.cpp
```diff
@@ -0,0 +1,119 @@
...
+    }
+    vector<vector<blackhorse::pt>> polygons2;
+    for (auto i : polygons) {
+        vector<blackhorse::pt> t;
+        for (auto j : i)
+            t.push_back({j.x, j.y});
+        polygons2.push_back(t);
+    }
+    auto val1 = polyUnion(polygons);
+    auto val2 = blackhorse::polygon_union(polygons2.data(), polygons2.size());
```
simonlindholm: Another way of testing this would be to approximate the area by checking "is (x,y) in any of the polygons" for regularly spaced x,y in a grid. Would be nice to remove the assumption about the underlying blackhorse impl being correct.

content/geometry/PolygonUnionArea.h
```diff
@@ -0,0 +1,49 @@
...
+					if (min(sc, sd) < 0)
+						segs.emplace_back(sa / (sa - sb), sgn(sc - sd));
+				} else if (!sc && !sd && j<i && sgn((B-A).dot(D-C))>0){
+					segs.emplace_back(rat2(C - A, B - A), 1);
+					segs.emplace_back(rat2(D - A, B - A), -1);
+				}
+			}
+		}
+		sort(all(segs));
+		trav(j,segs) j.first = min(max(j.first, 0.0), 1.0);
```
simonlindholm: rename `j` to something else since it's not an index

content/geometry/PolygonUnionArea.h
simonlindholm: `sgn(x) > 0` is the same as `x > 0`, right? or is the idea that epsilon checks can be added to `sgn` and `sideOf`? That should be mentioned if so

content/geometry/PolygonUnionArea.h
```diff
@@ -0,0 +1,49 @@
...
+ * Time: $O(N^2)$, where $N$ is the total number of points
+ * Usage:
+ */
+#pragma once
+
+#include "Point.h"
+#include "sideOf.h"
+
+typedef Point<double> P;
+double rat2(P a, P b) { return sgn(b.x) ? a.x/b.x : a.y/b.y; }
```
simonlindholm: (this function is super confusing to me btw)

fuzz-tests/geometry/PolygonUnionArea.cpp
```diff
@@ -0,0 +1,119 @@
...
+    return ret / 2;
+}
+} // namespace blackhorse
+
+void testRandom(int n, int numPts = 10, int lim = 5) {
+    vector<vector<P>> polygons;
+    for (int i = 0; i < n; i++) {
+        vector<P> pts;
+        for (int j = 0; j < numPts; j++) {
+            pts.push_back(P(rand() % lim, rand() % lim));
```
simonlindholm: `- lim/2` to test negative coordinates as well

content/geometry/PolygonUnionArea.h
```diff
@@ -0,0 +1,49 @@
+/**
+ * Author: black_horse2014, chilli
+ * Date: 2019-10-29
+ * License: CC0
+ * Source:
```
simonlindholm: Where did you get this from? Just to double-check that it's indeed fine with CC0.

content/geometry/PolygonUnionArea.h
```diff
@@ -0,0 +1,49 @@
...
+ * Description: Calculates the area of the union of $n$ polygons (not necessarily convex).
+ * Status: Submitted on ECNA 2017 Problem A.
```
simonlindholm: , fuzz-tested

content/geometry/PolygonUnionArea.h
Chillee: Honestly, I don't really understand what this function does either... It has to do with the case when  2 line segments have infinite intersection points. Kind of icky.

content/geometry/PolygonUnionArea.h
Chillee: Hmm, accidentally left that in here. Unfortunately, I don't know where the original source is from. The implementation I took the most from is: https://codeforces.com/gym/101673/submission/50481926 (blackhorse_2014's submission on that problem).

Considering that many of the implementations (the short ones at least) used the same conceptual implementation that's very similar, I suspect there's a single original source. Unfortunately, it seems to be fairly difficult to find.

content/geometry/PolygonUnionArea.h
Chillee: there's not actually that many comparisons on doubles. There's the `sideOf` comparisons, the comparison in `rat`, and one extra comparison in the conditional. I think that's a reasonable condition.

Yeah I'll add some notes about epsilons.

content/geometry/PolygonUnionArea.h
Chillee: Unfortunately, that does add 2 lines :/, since the `poly[i]` and `poly[j]` are used in different places.

content/geometry/PolygonUnionArea.h
```diff
@@ -0,0 +1,50 @@
...
+				} else if (!sc && !sd && j<i && sgn((B-A).dot(D-C))>0){
+					segs.emplace_back(rat(C - A, B - A), 1);
+					segs.emplace_back(rat(D - A, B - A), -1);
+				}
+			}
+		}
+		sort(all(segs));
+		trav(s,segs) s.first = min(max(s.first, 0.0), 1.0);
+		double sum = 0;
+		int cnt = segs[0].second;
```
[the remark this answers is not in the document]
Chillee: I would prefer not to use C++17 features that C++11 happens to accept. IIRC, not all GCC versions that implement the C++11 standard will allow C++17 features with a warning. So whether this works or not is at the mercy of what version of GCC the contest organizers are using.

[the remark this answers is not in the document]
Chillee: I'm also worried about the second one due to potential numerical precision issues.

content/geometry/PolygonUnion.h
```diff
@@ -0,0 +1,51 @@
...
+		double sum = 0;
+		int cnt = segs[0].second;
+		rep(j,1,sz(segs)) {
+			if (!cnt) sum += segs[j].first - segs[j - 1].first;
+			cnt += segs[j].second;
+		}
+		ret += A.cross(B) * sum;
+	}
+	return ret / 2;
+}
```
simonlindholm: duplicate file: you have both PolygonUnion.h and PolygonUnionArea.h

simonlindholm:
Looks like the algorithm is fine wrt float imprecision; the apparent asymmetry was my not reading the code closely enough. Not sure why epsilons were used in the first place...

Still missing Source and License, I'll merge with that fixed.

Chillee:
Unfortunately, I took it from a Gym submission on CF. I'll message him on CF, but not sure he'll respond.

Chillee:
Unfortunately he read the message but hasn't responded.

PR #135, opened 2019-10-31
Added circle polygon intersection

Chillee:
Not fully fleshed out yet.

### DIFF
--- a/content/geometry/CirclePolygonIntersection.h
+++ b/content/geometry/CirclePolygonIntersection.h
@@ -14,28 +14,21 @@
 #include "../../content/geometry/Point.h"
 
 typedef Point<double> P;
-double areaCT(P pa, P pb, double r) {
-	if (pa.dist() < pb.dist()) swap(pa, pb);
-	if (sgn(pb.dist()) == 0) return 0;
-	double a = pb.dist(), b = pa.dist(), c = (pb - pa).dist();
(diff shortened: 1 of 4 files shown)


Chillee:
I also found another implementation here that seems like it might be significantly shorter :thinking: https://github.com/spaghetti-source/algorithm/blob/master/geometry/_geom.cc#L744

Chillee:
I switched over to the other implementation.

1. It's significantly shorter.
2. I think it's more numerically accurate. It's kind of hard for me to reason about the actual numerical stability, but with no long doubles, this new implementation + the original differed by >1e-7 on cases within 10 of the origin. If I added long doubles to the original, they're within 1e-8. However, adding long doubles to the new implementation doesn't make their difference any smaller.

From that, I conclude that the new implementation is more accurate.

content/geometry/CirclePolygonIntersection.h
```diff
@@ -0,0 +1,33 @@
+/**
+ * Author: chilli, Takanori MAEHARA
+ * Date: 2019-10-31
+ * License: CC0
+ * Source: https://github.com/spaghetti-source/algorithm/blob/master/geometry/_geom.cc#L744
+ * Description: Returns the area of the intersection of a circle with a
+ * polygon.
```
simonlindholm: "(ccw) polygon"

PR #136, opened 2019-11-01
Added circle circle tangents

Chillee:
Unfortunately, it's significantly longer, but it's also significantly more general. Personally I've found circle circle tangents to pop up in more problems than circle point tangents.


content/geometry/circleTangents.h
```diff
@@ -1,27 +1,32 @@
 /**
- * Author: Ulf Lundstrom
- * Date: 2009-04-06
+ * Author: Victor Lecomte, chilli
+ * Date: 2019-10-31
  * License: CC0
  * Source:
```
simonlindholm: add source

content/geometry/circleTangents.h
```diff
@@ -1,27 +1,32 @@
 /**
```
simonlindholm: you can rename the file to CircleTangents.h

content/geometry/circleTangents.h
```diff
@@ -1,27 +1,32 @@
...
  * License: CC0
  * Source:
- * Description:\\
-\begin{minipage}{75mm}
-Returns a pair of the two points on the circle with radius r centered around c whos tangent lines intersect p. If p lies within the circle NaN-points are returned. P is intended to be Point<double>. The first point is the one to the right as seen from the p towards c.
-\end{minipage}
-\begin{minipage}{15mm}
-\includegraphics[width=\textwidth]{content/geometry/circleTangents}
-\end{minipage}
+ * Description: Returns the external tangents of two circles, where P1 and P2
```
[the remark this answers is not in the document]
Chillee: yeah

content/geometry/circleTangents.h
```diff
@@ -1,27 +1,32 @@
...
+ * are the tangency points of circle 1 and 2 respectively. Can return 0, 1, or
+ * 2 tangents. If the circles are the same, returns 0 tangents. If there is
```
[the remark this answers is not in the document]
Chillee: Hmm, I believe it returns 0 tangents if two circles are nested inside each other (which doesn't necessarily require the centers to be the same) or if the circles are identical (thus infinite tangents).

content/geometry/circleTangents.h
```diff
@@ -1,27 +1,32 @@
...
+ * only 1 tangent, then the circles are tangent to each other at some point P,
+ * and the tangent line can be found as the perpendicular line c1 -> P passing
```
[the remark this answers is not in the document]
Chillee: yes

fuzz-tests/utilities/randGeo.h
```diff
@@ -0,0 +1,12 @@
+#pragma once
+
+#include <bits/stdc++.h>
+#include "../../content/geometry/Point.h"
+
+#define all(x) begin(x), end(x)
+typedef long long ll;
+using namespace std;
+
+Point<double> randPt(int lim) {
```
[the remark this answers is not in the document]
Chillee: Let's keep them separate for now. If we have more geo utilities later we can put them all in the same file.

fuzz-tests/geometry/CircleTangents.cpp
```diff
@@ -0,0 +1,48 @@
...
+
+#define rep(i, a, b) for (int i = a; i < (b); ++i)
+#define trav(a, x) for (auto &a : x)
+#define all(x) begin(x), end(x)
+#define sz(x) (int)(x).size()
+typedef long long ll;
+typedef pair<int, int> pii;
+typedef vector<int> vi;
+
+#include "../../content/geometry/circleTangents.h"
```
simonlindholm: Update this reference and the one in chapter.tex

content/geometry/CircleTangents.h
```diff
@@ -0,0 +1,32 @@
...
+ * Date: 2019-10-31
+ * License: CC0
+ * Source: https://vlecomte.github.io/cp-geo.pdf
+ * Description: Returns the external tangents of two circles, where P1 and P2
+ * are the tangency points of circle 1 and 2 respectively. Can return 0, 1, or
+ * 2 tangents. If the circles have no tangents/are identical, returns 0 tangents. If there is
+ * only 1 tangent, then the circles are tangent to each other at some point P,
+ * and the tangent line can be found as the perpendicular line c1 -> P passing
+ * through P. Internal tangents can be found by negating r2. To find the
+ * tangents of a circle with a point c2 set r2 to 0.
```
simonlindholm: I tried to do some wordsmithing here. How does the following sound to you?
```
Description: Finds the external tangents of two circles, or internal if r2 is
negated. Can return 0, 1, or 2 tangents – 0 if one circle contains the other (or
overlaps it, in the internal case); 1 if the circles are tangent to each other (in
which case .first = .second and the tangent line is perpendicular to the line
between the centers). To find the tangents of a circle with a point c2 set r2
to 0. .first and .second give the tangency points at circle 1 and 2 respectively.
```

It tries to fix a few points:
* `can be found as the perpendicular line c1 -> P` needs "of the line", but also isn't totally correct if c1 = P, so I changed this to the perpendicular of the line c1 - c2
* clarify when the 0/1 tangent cases occur

Chillee: Seems a lot better. I might move `To find the tangents of a circle with a point c2 set r2
to 0.` to the end. Also, I want to clarify that if there are infinite tangents, it also returns 0.

simonlindholm:
Updated the description. Not sure if the identical circles case was worth mentioning, honestly, but oh well...

PR #138, opened 2019-11-03
Added Manhattan MST

Chillee:
The main API decision to make here is whether we want to include MST within the implementation or to just return the set of candidate edges.

If we want to include the MST within the implementation, then I think we should include the MST code inside the notebook elsewhere and call it from here.

### DIFF
--- a/content/geometry/ManhattanMST.h
+++ b/content/geometry/ManhattanMST.h
@@ -23,9 +23,9 @@ vector<array<int, 3>> manhattanMST(vector<P> ps) {
 			for (auto it = sweep.lower_bound(-ps[i].y);
 						it != sweep.end(); sweep.erase(it++)) {
 				int j = it->second;
-				P d = ps[j] - ps[i];
-				if (d.y < d.x) break;
-				edges.push_back({abs(d.y) + abs(d.x), i, j});
+				P d = ps[i] - ps[j];
+				if (d.y > d.x) break;
+				edges.push_back({d.y + d.x, i, j});
 			}
 			sweep[-ps[i].y] = i;
 		}
--- a/content/geometry/ManhattanMST.h
+++ b/content/geometry/ManhattanMST.h
@@ -15,24 +15,22 @@ vector<array<int, 3>> manhattanMST(vector<P> ps) {
 	vi id(sz(ps));
 	iota(all(id), 0);
 	vector<array<int, 3>> edges;
-	rep(s,0,2) {
-		rep(t,0,2) {
-			sort(all(id), [&](int i, int j) {
-			     return (ps[i]-ps[j]).x < (ps[j]-ps[i]).y;});
-			map<int, int> sweep;
-			trav(i,id) {
-				for (auto it = sweep.lower_bound(-ps[i].y);
-					        it != sweep.end(); sweep.erase(it++)) {
-					int j = it->second;
-					auto d = ps[j] - ps[i];
-					if (d.y < d.x) break;
-					edges.push_back({abs(d.y) + abs(d.x), i, j});
-				}
-				sweep[-ps[i].y] = i;
+	rep(k,0,4) {
+		sort(all(id), [&](int i, int j) {
+				return (ps[i]-ps[j]).x < (ps[j]-ps[i]).y;});
+		map<int, int> sweep;
+		trav(i,id) {
+			for (auto it = sweep.lower_bound(-ps[i].y);
+						it != sweep.end(); sweep.erase(it++)) {
+				int j = it->second;
+				auto d = ps[j] - ps[i];
+				if (d.y < d.x) break;
+				edges.push_back({abs(d.y) + abs(d.x), i, j});
 			}
-			trav(p,ps) swap(p.y, p.x);
+			sweep[-ps[i].y] = i;
 		}
-		trav(p,ps) p.x = -p.x;
+		if (k & 1) trav(p,ps) p.x = -p.x;
+		else trav(p,ps) swap(p.x, p.y);
 	}
 	return edges;
 }
--- a/content/geometry/ManhattanMST.h
+++ b/content/geometry/ManhattanMST.h
@@ -3,12 +3,10 @@
  * Date: 2019-11-02
  * License: CC0
  * Source: https://github.com/spaghetti-source/algorithm/blob/master/geometry/rectilinear_mst.cc
- * Description: Given N points, returns up to 8*N edges, which are guaranteed
- * to contain the Manhattan MST. The Manhattan MST of a set of points is the
- * MST where the weight of the edge between any pair of points is the Manhattan
- * distance between the points. Edges are in the form (distance, src, dst). Use
- * a standard MST algorithm on the result to find the final MST. Modify `edges`
- * and `sweep` to support doubles.
+ * Description: Given N points, returns up to 4*N edges, which are guaranteed
+ * to contain a minimum spanning tree for the graph with edge weights w(p, q) =
+ * |p.x - q.x| + |p.y - q.y|. Edges are in the form (distance, src, dst). Use a
+ * standard MST algorithm on the result to find the final MST.
  * Time: O(N log N)
  * Status: Fuzz-tested
  */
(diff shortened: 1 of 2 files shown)


content/geometry/ManhattanMST.h
```diff
@@ -0,0 +1,36 @@
+/**
+ * Author: chilli, Takanori MAEHARA
+ * Date: 2019-11-02
+ * License: CC0
+ * Source: https://github.com/spaghetti-source/algorithm/blob/master/geometry/rectilinear_mst.cc
+ * Description: Returns up to 8*N edges, which are guaranteed to contain the
```
simonlindholm: Add that it's for the 2d case with integer coordinates, and maybe explain that Manhattan MST means.

content/geometry/ManhattanMST.h
```diff
@@ -0,0 +1,36 @@
...
+	iota(all(id), 0);
+	vector<array<int, 3>> edges;
+	rep(s,0,2) {
+		rep(t,0,2) {
+			sort(all(id), [&](int i, int j) {
+					return (ps[i]-ps[j]).x < (ps[j]-ps[i]).y;});
+			map<double, int> sweep;
+			for (int i: id) {
+				for (auto it = sweep.lower_bound(-ps[i].y);
+									it != sweep.end(); sweep.erase(it++)) {
```
simonlindholm: nit: align using spaces

content/geometry/ManhattanMST.h
```diff
@@ -0,0 +1,36 @@
...
+					int j = it->second;
+					auto d = ps[j] - ps[i];
+					if (d.y < d.x) break;
+					edges.push_back({abs(d.y) + abs(d.x), i, j});
+				}
+				sweep[-ps[i].y] = i;
+			}
+			for (auto &p: ps) p = P(p.y, p.x);
```
simonlindholm: `trav(p, ps)`

content/geometry/ManhattanMST.h
```diff
@@ -0,0 +1,36 @@
...
+typedef Point<int> P;
+vector<array<int, 3>> manhattanMST(vector<P> ps) {
+	vector<int> id(sz(ps));
+	iota(all(id), 0);
+	vector<array<int, 3>> edges;
+	rep(s,0,2) {
+		rep(t,0,2) {
+			sort(all(id), [&](int i, int j) {
+					return (ps[i]-ps[j]).x < (ps[j]-ps[i]).y;});
+			map<double, int> sweep;
```
simonlindholm: why double?

content/geometry/ManhattanMST.h
Chillee: Actually, I think I'm gonna modify it to work with ints/doubles - I think it's fairly reasonable to call this on doubles.

Chillee: Actually, nvm, that's pretty annoying...

simonlindholm: Ugh, doubles, now I need to sit down and analyze this for precision issues... I think it would be fine to use ints everywhere and leave it to users to change to doubles if needed; I don't think I've ever seen a problem which used Manhattan metric without integer coordinates.

fuzz-tests/geometry/ManhattanMST.cpp
```diff
@@ -0,0 +1,56 @@
...
+  return cost;
+}
+
+signed main() {
+    for (int t=0; t<100; t++) {
+        int num_pts = 1000;
+        vector<P> pts;
+        for (int i = 0; i < num_pts; ++i) {
+            double x = rand() % 300;
+            double y = rand() % 300;
```
simonlindholm: negative numbers don't hurt

Chillee:
I actually switched back to integers so no floats necessary :^)

Besides, the only floats that would be used are absolute differences in the floating point values and adding them together, so I don't think we need to worry about it much.

simonlindholm:
Well, there's still a note in the description about floats. If you remove that note I can remove the scary-looking label. :)

(My worries re floats here are if float imprecision caused segments to be skipped based on a "there are better candidates" argument, and those skip-in-favor-of ended up circular. I'd need to understand the algorithm to determine whether that can actually be the case.)

content/geometry/ManhattanMST.h
```diff
@@ -0,0 +1,40 @@
+/**
+ * Author: chilli, Takanori MAEHARA
+ * Date: 2019-11-02
+ * License: CC0
+ * Source: https://github.com/spaghetti-source/algorithm/blob/master/geometry/rectilinear_mst.cc
+ * Description: Given N points, returns up to 8*N edges, which are guaranteed
```
simonlindholm: `4*N` (doesn't hurt to assert this in the test)

content/geometry/ManhattanMST.h
```diff
@@ -0,0 +1,40 @@
...
+ * to contain the Manhattan MST. The Manhattan MST of a set of points is the
+ * MST where the weight of the edge between any pair of points is the Manhattan
+ * distance between the points. Edges are in the form (distance, src, dst). Use
+ * a standard MST algorithm on the result to find the final MST. Modify `edges`
+ * and `sweep` to support doubles.
```
simonlindholm: This comment seems fairly useless, it's easy enough to figure out on your own

content/geometry/ManhattanMST.h
```diff
@@ -0,0 +1,40 @@
...
+					int j = it->second;
+					auto d = ps[j] - ps[i];
+					if (d.y < d.x) break;
+					edges.push_back({abs(d.y) + abs(d.x), i, j});
+				}
+				sweep[-ps[i].y] = i;
+			}
+			trav(p,ps) swap(p.y, p.x);
+		}
+		trav(p,ps) p.x = -p.x;
```
simonlindholm: Possible shortening: `rep(k,0,4)` for the outermost two loops, then
```cpp
if (k & 1) trav(p,ps) p.x = -p.x;
else trav(p,ps) swap(p.x, p.y);
```
(if there's a generator for the symmetry group in question we could get away without the if statement)

content/geometry/ManhattanMST.h
```diff
@@ -0,0 +1,40 @@
...
+vector<array<int, 3>> manhattanMST(vector<P> ps) {
+	vi id(sz(ps));
+	iota(all(id), 0);
+	vector<array<int, 3>> edges;
+	rep(s,0,2) {
+		rep(t,0,2) {
+			sort(all(id), [&](int i, int j) {
+			     return (ps[i]-ps[j]).x < (ps[j]-ps[i]).y;});
+			map<int, int> sweep;
+			for (int i: id) {
```
simonlindholm: space around `:`

content/geometry/ManhattanMST.h
```diff
@@ -0,0 +1,40 @@
+/**
+ * Author: chilli, Takanori MAEHARA
+ * Date: 2019-11-02
+ * License: CC0
+ * Source: https://github.com/spaghetti-source/algorithm/blob/master/geometry/rectilinear_mst.cc
+ * Description: Given N points, returns up to 8*N edges, which are guaranteed
+ * to contain the Manhattan MST. The Manhattan MST of a set of points is the
+ * MST where the weight of the edge between any pair of points is the Manhattan
+ * distance between the points. Edges are in the form (distance, src, dst). Use
```
simonlindholm: "...guaranteed to contain a minimum spanning tree for the graph with edge weights `w(p, q) = |p.x - q.x| + |p.y - q.y|`."? (the current wording is a bit verbose)

content/geometry/ManhattanMST.h
Chillee: Saves only one character but 2 lines, so ok :)

content/geometry/ManhattanMST.h
Chillee: just replaced this with trav.

Chillee:
Alright, shortened the documentation.

Technically I removed the reference to floats :^). If you still wanna check it out for float correctness, feel free. I'm pretty sure it'll be fine though - the points are never modified and the only operation made to them that could introduce FPE in the algorithm is `auto d = ps[j] - ps[i]`.

content/geometry/ManhattanMST.h
```diff
@@ -0,0 +1,39 @@
...
+			sort(all(id), [&](int i, int j) {
+					return (ps[i]-ps[j]).x < (ps[j]-ps[i]).y;});
+			map<T, int> sweep;
+			for (int i: id) {
+				for (auto it = sweep.lower_bound(-ps[i].y);
+					        it != sweep.end(); sweep.erase(it++)) {
+					int j = it->second;
+					auto d = ps[j] - ps[i];
+					if (d.y < d.x) break;
+					edges.push_back({abs(d.y) + abs(d.x), {i, j}});
```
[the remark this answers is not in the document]
Chillee: yeah that's true.

PR #139, opened 2019-11-03
Added circle circle intersection area

Chillee:
No tests - sanity checked it.


content/geometry/CircleCircleArea.h
```diff
@@ -0,0 +1,21 @@
...
+  if (cr < dr) swap(c, d);
+  auto A = [&](double r, double h) {
+    return r*r*acos(h/r)-h*sqrt(r*r-h*h);
+  };
+  auto l = (c - d).dist(), a = (l*l + cr*cr - dr*dr)/(2*l);
+  if (sgn(l - cr - dr) >= 0) return 0; // far away
+  if (sgn(l - cr + dr) <= 0) return M_PI*dr*dr;
+  if (sgn(l - cr) >= 0) return A(cr, a) + A(dr, l-a);
+  else return A(cr, a) + M_PI*dr*dr - A(dr, a-l);
+}
```
simonlindholm: Tabs for indentation (I should change to real 2-space indentation, I know...)

content/geometry/CircleCircleArea.h
```diff
@@ -0,0 +1,21 @@
...
+template<class P>
+double intersection_area(P c, double cr, P d, double dr) {
+  if (cr < dr) swap(c, d);
+  auto A = [&](double r, double h) {
+    return r*r*acos(h/r)-h*sqrt(r*r-h*h);
+  };
+  auto l = (c - d).dist(), a = (l*l + cr*cr - dr*dr)/(2*l);
+  if (sgn(l - cr - dr) >= 0) return 0; // far away
+  if (sgn(l - cr + dr) <= 0) return M_PI*dr*dr;
+  if (sgn(l - cr) >= 0) return A(cr, a) + A(dr, l-a);
```
simonlindholm: I'm not a fan of the `sgn`s here, there's no reason why this should be dependent on epsilons.

content/geometry/CircleCircleArea.h
```diff
@@ -0,0 +1,21 @@
...
+ * Date: 2019-11-03
+ * License: CC0
+ * Source: https://github.com/spaghetti-source/algorithm/blob/master/geometry/_geom.cc#L729
+ * Description: Calculates the area of the intersection of 2 circles
+ * Status:
+ */
+
+template<class P>
+double intersection_area(P c, double cr, P d, double dr) {
+  if (cr < dr) swap(c, d);
```
simonlindholm: swap cr/dr as well

content/geometry/CircleCircleArea.h
Chillee: no I like the tabs actually. I just keep on forgetting...

content/geometry/CircleCircleArea.h
Chillee: Arguably if you pass in two circles that don't touch, you want it to return absolute 0.

I do think it fails gracefully though, so I guess it's fine to remove it.

content/geometry/CircleCircleArea.h
simonlindholm: Did you accidentally revert the spaces -> tabs change?

content/geometry/CircleCircleArea.h
```diff
@@ -0,0 +1,21 @@
...
+ * Source: https://github.com/spaghetti-source/algorithm/blob/master/geometry/_geom.cc#L729
+ * Description: Calculates the area of the intersection of 2 circles
+ * Status:
+ */
+
+template<class P>
+double circleCircleArea(P c, double cr, P d, double dr) {
+	if (cr < dr) swap(c, d), swap(cr, dr);
+	auto A = [&](double r, double h) {
+		return r*r*acos(h/r)-h*sqrt(r*r-h*h);
```
simonlindholm: The removal of the epsilons now means that r < h can happen due to numerical precision, I suspect (though I haven't tested). `h = min(h, r);` may be reasonable? `h < 0` can likely also happen but doesn't seem like a problem.

content/geometry/CircleCircleArea.h
```diff
@@ -0,0 +1,21 @@
...
+ * Date: 2019-11-03
+ * License: CC0
+ * Source: https://github.com/spaghetti-source/algorithm/blob/master/geometry/_geom.cc#L729
+ * Description: Calculates the area of the intersection of 2 circles
+ * Status:
+ */
+
+template<class P>
+double circleCircleArea(P c, double cr, P d, double dr) {
+	if (cr < dr) swap(c, d), swap(cr, dr);
```
simonlindholm: hm, my suggestion to add `swap(c, d)` was silly, only the distance between them is used and that's symmetric

content/geometry/CircleCircleArea.h
```diff
@@ -0,0 +1,21 @@
...
+	auto A = [&](double r, double h) {
+		return r*r*acos(h/r)-h*sqrt(r*r-h*h);
+	};
+	auto l = (c - d).dist(), a = (l*l + cr*cr - dr*dr)/(2*l);
+	if (l - cr - dr >= 0) return 0; // far away
+	if (l - cr + dr <= 0) return M_PI*dr*dr;
+	if (l - cr >= 0) return A(cr, a) + A(dr, l-a);
```
simonlindholm: these inequalities can be simplified now without the epsilons

content/geometry/CircleCircleArea.h
```diff
@@ -0,0 +1,21 @@
...
+ * Status:
+ */
+
+template<class P>
+double circleCircleArea(P c, double cr, P d, double dr) {
+	if (cr < dr) swap(c, d), swap(cr, dr);
+	auto A = [&](double r, double h) {
+		return r*r*acos(h/r)-h*sqrt(r*r-h*h);
+	};
+	auto l = (c - d).dist(), a = (l*l + cr*cr - dr*dr)/(2*l);
```
simonlindholm: `double` is clearer than `auto`

PR #140, opened 2019-11-04
Added fast input

Chillee:
I remember we had some discussions about the most ergonomic API, feel free to suggest changes.

Other than that, there's 2 main decisions that we want to consider.

1. Using `fread` instead of `getchar_unlocked`. `fread` is about 2x faster, for a total of 6x faster than cin.
2. Augment this implementation to read in floats as well.

First one seems more worth it. It's fairly easy to read in floats if needed (by reading both parts and doing some manual division). In addition, if you're ever in the situation where you need fast IO you might as well use the fastest IO you can get.

### DIFF
--- a/content/various/FastInput.h
+++ b/content/various/FastInput.h
@@ -1,0 +1,33 @@
+/**
+ * Author: chilli
+ * License: CC0
+ * Source:
+ * Description: Reads in arbitrary amounts of integral values.
+ * Status: tested on SPOJ INTEST
+ * Time: About 3x as fast as cin.
+ * Usage: int a, b[5]; ll c;
+ * read_int(a, b[0], c);
+ */
+#pragma once
+
+struct GC {
+    char buf[1 << 16 | 1];
+    int bc = 0, be = 0;
+    char operator()() {
+        if (bc >= be) {
+            be = fread(buf, 1, sizeof(buf) - 1, stdin);
+            buf[be] = bc = 0;
+        }
+        return buf[bc++];
+    }
+} gc;
+void read_int() {}
+template <class T, class... S>
+inline void read_int(T &a, S &... b) {
+	char c, s = 1;
+	while (isspace(c = gc()));
+	if (c == '-') s = -1, c = gc();
+	for (a = c - '0'; isdigit(c = gc()); a = a * 10 + c - '0');
+	a *= s;
+	read_int(b...);
+}
--- a/content/various/FastInput.h
+++ b/content/various/FastInput.h
@@ -2,32 +2,29 @@
  * Author: chilli
  * License: CC0
  * Source:
- * Description: Reads in arbitrary amounts of integral values.
+ * Description: Returns an integer. Usage requires your program to pipe in input from file.
  * Status: tested on SPOJ INTEST
- * Time: About 5x as fast as cin.
- * Usage: int a, b[5]; ll c;
- * read_int(a, b[0], c);
+ * Time: About 5x as fast as cin/scanf.
+ * Usage: ./a.out < input.txt
  */
 #pragma once
 
 struct GC {
-    char buf[1 << 16 | 1];
-    int bc = 0, be = 0;
-    char operator()() {
-        if (bc >= be) {
-            be = fread(buf, 1, sizeof(buf) - 1, stdin);
-            buf[be] = bc = 0;
-        }
-        return buf[bc++];
-    }
+	char buf[1 << 16 | 1];
+	int bc = 0, be = 0;
+	char operator()() {
+		if (bc >= be) {
+			be = fread(buf, 1, sizeof(buf) - 1, stdin);
+			buf[be] = bc = 0;
+		}
+		return buf[bc++];
+	}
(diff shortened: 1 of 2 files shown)


content/various/FastInput.h
```diff
@@ -0,0 +1,33 @@
...
+    char buf[1 << 16 | 1];
+    int bc = 0, be = 0;
+    char operator()() {
+        if (bc >= be) {
+            be = fread(buf, 1, sizeof(buf) - 1, stdin);
+            buf[be] = bc = 0;
+        }
+        return buf[bc++];
+    }
+} gc;
```
simonlindholm: IO/io may be a better name, I confuse GC with garbage collection

content/various/FastInput.h
```diff
@@ -0,0 +1,33 @@
...
+
+struct GC {
+    char buf[1 << 16 | 1];
+    int bc = 0, be = 0;
+    char operator()() {
+        if (bc >= be) {
+            be = fread(buf, 1, sizeof(buf) - 1, stdin);
+            buf[be] = bc = 0;
+        }
+        return buf[bc++];
```
simonlindholm: add a comment saying this returns 0 on eof, maybe?

content/various/FastInput.h
```diff
@@ -0,0 +1,33 @@
...
+void read_int() {}
+template <class T, class... S>
+inline void read_int(T &a, S &... b) {
+	char c, s = 1;
+	while (isspace(c = gc()));
+	if (c == '-') s = -1, c = gc();
+	for (a = c - '0'; isdigit(c = gc()); a = a * 10 + c - '0');
+	a *= s;
+	read_int(b...);
+}
```
simonlindholm: I would prefer the API `int read_int();`, mostly because the implementation becomes shorter. (Let's you do `if (c == '-') return -read_int();`, e.g.) It also helps the compiler understand that pointer to the return value doesn't escape, but you've added inline so that's not really an argument.

It might be a good idea to replace `isspace(x)` by `x < 40` and similarly for `isdigit`, since those functions use the current locale and may be slow (?).

content/various/FastInput.h
Chillee: I use `gc` to stand for `getChar`.

content/various/FastInput.h
Chillee: I think one of the benefits of doing it this way is that we can read long longs/ints depending solely on what we pass in as arguments.

Lemme think about it a bit more. As book code, it doesn't matter as much how ergonomic it is, but I think it's still nice.

Chillee: Interestingly, replacing `isspace` with the boolean comparison makes a substantial difference (~20%). Replacing `isdigit` with the boolean comparison doesn't make much of a difference though.

Chillee:
Updated - added some more description about how to use.

The first part is definitely a good idea (it's confusing if you replace your IO with this and then all of a sudden you can't test through interactive terminal).

content/various/FastInput.h
```diff
@@ -0,0 +1,32 @@
...
+ */
+#pragma once
+
+struct GC {
+	char buf[1 << 16 | 1];
+	int bc = 0, be = 0;
+	char operator()() {
+		if (bc >= be) {
+			be = fread(buf, 1, sizeof(buf) - 1, stdin);
+			buf[be] = bc = 0;
```
simonlindholm: The whole scheme with adding +1 to the buf size and zero-terminating this buffer is just to deal with case where fread returns 0, right? how about
```
buf[0] = bc = 0;
be = fread(buf, 1, sizeof(buf), stdin);
```
and making buf have size 1 << 16?

content/various/FastInput.h
```diff
@@ -0,0 +1,32 @@
...
+		}
+		return buf[bc++]; // returns 0 on EOF
+	}
+} gc;
+inline int read_int() {
+	static char c;
```
simonlindholm: why is `c` static?

content/various/FastInput.h
```diff
@@ -0,0 +1,32 @@
...
+	while ((c = gc()) < 40);
+	if (c == '-') return -read_int();
+	int a = c - '0';
+	for (; isdigit(c = gc()); a = a * 10 + c - '0');
```
simonlindholm: use a while loop for this instead of for

content/various/FastInput.h
Chillee: The point is so that `gc()` returns 0 on EOF. I don't think that happens with this right?

simonlindholm: So in my version, if fread returns zero, we return the value of `buf[0]`, which is set to zero beforehand. This assumes that fread doesn't write past the size it returns, which I *think* should be true per spec, and it's hard to imagine what garbage it would write there if not, but I didn't read super closely or test in practice.

Chillee: yeah you're right.

Chillee:
Updated - I tested it on some examples and it seems to work fine.

PR #144, opened 2019-11-08
Added slightly more description for modsqrt

Chillee:
I think it's a good idea to give a mathematical description of what it computes. Otherwise, variable names could be slightly confusing.


simonlindholm:
Good call. I expanded it a bit further.

PR #145, opened 2019-11-08
Added Discrete log

Chillee:
I'm fairly surprised we don't have this, tbh.

We might want to add instructions for how to extend this to arbitary a and m. I don't remember how, but I think it's possible.

### DIFF
--- a/content/number-theory/ModLog.h
+++ b/content/number-theory/ModLog.h
@@ -1,26 +1,22 @@
-
 /**
  * Author: Andrei Navumenka, chilli
  * Date: 2019-11-08
  * License: Unlicense
  * Source: https://github.com/indy256/codelibrary/blob/master/cpp/numbertheory/discrete_log.cpp
  * Description: Calculates the discrete log. Returns an x s.t. a^x = b mod m. a and m must be coprime.
- * Time: $O(n^{1/2})$
+ * Time: $O(m^{1/2})$
  * Status: tested for all values under a 1000 for a,x, and m.
  */
 #pragma once
 
 ll modLog(ll a, ll b, ll m) {
 	assert(__gcd(a, m) == 1);
-	ll n = (ll) sqrt(m) + 1, an = 1;
-	rep(i,0,n) an = an * a % m;
-	unordered_map<ll, ll> vals;
-	for (ll i = 1, x = an; i <= n; i++, x = x * an % m)
-		if (!vals.count(x)) vals[x] = i;
+	if (b == 1) return 0;
+	ll n = (ll) sqrt(m) + 1, e = 1;
+	rep(i,0,n) e = e * a % m;
+	unordered_map<ll, ll> f;
+	for (ll i=1, x = e; i<=n; i++, x = x*e%m) f.emplace(x, i);
 	for (ll i = 0, x = b; i <= n; i++, x = x * a % m)
-		if (vals.count(x)) {
-			ll res = vals[x] * n - i;
-			if (res < m) return res;
-		}
+		if (f.count(x)) return f[x] * n - i;
 	return -1;
 }
--- a/content/number-theory/ModLog.h
+++ b/content/number-theory/ModLog.h
@@ -3,20 +3,19 @@
  * Date: 2019-11-08
  * License: Unlicense
  * Source: https://github.com/indy256/codelibrary/blob/master/cpp/numbertheory/discrete_log.cpp
- * Description: Calculates the discrete log. Returns an x s.t. a^x = b mod m. a and m must be coprime.
- * Time: $O(m^{1/2})$
- * Status: tested for all values under a 1000 for a,x, and m.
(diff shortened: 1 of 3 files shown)


fuzz-tests/number-theory/ModLog.cpp
```diff
@@ -0,0 +1,32 @@
...
+
+#include "../../content/number-theory/ModMulLL.h"
+#include "../../content/number-theory/ModLog.h"
+
+uniform_int_distribution<ll> uni(1, 100);
+const int ITERS = 1e7;
+const int LIM = 1000;
+int main() {
+    rep(m,1,500) {
+        rep(a,1,m) {
```
simonlindholm: `rep(a,0,m)` just to cover the `a = 0, m = 1` case

content/number-theory/ModLog.h
```diff
@@ -0,0 +1,26 @@
+
+/**
+ * Author: Andrei Navumenka, chilli
+ * Date: 2019-11-08
+ * License: Unlicense
+ * Source: https://github.com/indy256/codelibrary/blob/master/cpp/numbertheory/discrete_log.cpp
+ * Description: Calculates the discrete log. Returns an x s.t. a^x = b mod m. a and m must be coprime.
+ * Time: $O(n^{1/2})$
+ * Status: tested for all values under a 1000 for a,x, and m.
```
simonlindholm: (the test says 500 but I guess you tested 1000)

content/number-theory/ModLog.h
```diff
@@ -0,0 +1,26 @@
...
+	assert(__gcd(a, m) == 1);
+	ll n = (ll) sqrt(m) + 1, an = 1;
+	rep(i,0,n) an = an * a % m;
+	unordered_map<ll, ll> vals;
+	for (ll i = 1, x = an; i <= n; i++, x = x * an % m)
+		if (!vals.count(x)) vals[x] = i;
+	for (ll i = 0, x = b; i <= n; i++, x = x * a % m)
+		if (vals.count(x)) {
+			ll res = vals[x] * n - i;
+			if (res < m) return res;
```
simonlindholm: I don't see the point of this if statement; checking `< m` isn't enough to get a unique answer. Maybe we should take the minimum of all `res` instead, or add a comment in the description to that effect.

content/number-theory/ModLog.h
```diff
@@ -0,0 +1,26 @@
+
+/**
+ * Author: Andrei Navumenka, chilli
+ * Date: 2019-11-08
+ * License: Unlicense
+ * Source: https://github.com/indy256/codelibrary/blob/master/cpp/numbertheory/discrete_log.cpp
+ * Description: Calculates the discrete log. Returns an x s.t. a^x = b mod m. a and m must be coprime.
+ * Time: $O(n^{1/2})$
```
simonlindholm: `m`

content/number-theory/ModLog.h
```diff
@@ -0,0 +1,26 @@
+
```
simonlindholm: remove blank line

content/number-theory/ModLog.h
```diff
@@ -0,0 +1,26 @@
...
+ */
+#pragma once
+
+ll modLog(ll a, ll b, ll m) {
+	assert(__gcd(a, m) == 1);
+	ll n = (ll) sqrt(m) + 1, an = 1;
+	rep(i,0,n) an = an * a % m;
+	unordered_map<ll, ll> vals;
+	for (ll i = 1, x = an; i <= n; i++, x = x * an % m)
+		if (!vals.count(x)) vals[x] = i;
```
simonlindholm: `vals.emplace(x, i);` to get rid of the double lookup (shorter as well!)

content/number-theory/ModLog.h
Chillee: I think it might be awkward if it returns a value more than `m`.

simonlindholm: But why? `m` is an unnatural thing to compare an exponent to, which lives mod phi(m). Taking the minimum seems like a good compromise to me -- it would be easy to explain and give a canonical output.

Chillee: Does the big step giant step algorithm necessarily enumerate through all possibilities for `x`?

fuzz-tests/number-theory/ModLog.cpp
Chillee: Unfortunately, it fails on `0^x = 1 mod 1` case.

Chillee: That's annoying :thinking:

content/number-theory/ModLog.h
simonlindholm: It should, and the `emplace` makes sure to only keep the first entry for each key, so we get the smallest value. Should clearly be tested though.

Chillee: I think I'd rather keep it as is - I don't think it's a big deal which value it returns.

simonlindholm:
Shortened a bit and made it return the smallest answer.

PR #147, opened 2019-11-12
Removed leftover gcd code

Chillee:


simonlindholm:
It's not left-over as much as it's telling you that you should `__gcd` if you didn't already know about it. We could move it to the description though, maybe?

content/number-theory/euclid.h
```diff
@@ -2,14 +2,12 @@
  * Author: Unknown
  * Date: 2002-09-15
  * Source: predates tinyKACTL
- * Description: Finds the Greatest Common Divisor to the integers $a$ and $b$.
- * Euclid also finds two integers $x$ and $y$, such that $ax+by=\gcd(a,b)$.
+ * Description: Finds two integers $x$ and $y$, such that $ax+by=\gcd(a,b)$. If
+ * you just need gcd, use the built in __gcd instead
```
simonlindholm:
```suggestion
 * you just need gcd, use the built in \texttt{\_\_gcd} instead.
```
(just guessing at the underscore, but I think latex will parse them as math otherwise?)

PR #148, opened 2019-11-13
Added MIT Math stuff

Chillee:
Same stuff that was in the MIT 2019 notebook, just copy pasted. Added bernoulli numbers and labeled unrooted trees. Looks like this:

### DIFF
--- a/content/combinatorial/chapter.tex
+++ b/content/combinatorial/chapter.tex
@@ -39,6 +39,21 @@
 		\kactlimport{multinomial.h}
 
 \section{General purpose numbers}
+	\subsection{Bernoulli numbers}
+		EGF of Bernoulli numbers is $B(t)=\frac{t}{e^t-1}$ (FFT-able).
+		$B[0,\ldots] = [1, -\frac{1}{2}, \frac{1}{6}, 0, -\frac{1}{30}, 0, \frac{1}{42}, \ldots]$
+
+		Sums of powers:
+		\small
+		\[ \sum_{i=1}^n n^m = \frac{1}{m+1} \sum_{k=0}^m \binom{m+1}{k} B_k (n+1)^{m+1-k} \]
+		\normalsize
+
+		Euler-Maclaurin formula for infinite sums:
+		\small
+		\[ \sum_{i=m}^{\infty} f(i) = \int_m^\infty f(x) dx - \sum_{k=1}^\infty \frac{B_k}{k!}f^{(k-1)}(m) \]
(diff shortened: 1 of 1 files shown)


content/combinatorial/chapter.tex
```diff
@@ -39,6 +39,21 @@ \section{Partitions and subsets}
 		\kactlimport{multinomial.h}
 
 \section{General purpose numbers}
+	\subsection{Bernoulli numbers}
+		EGF of Bernoulli numbers is $B(t)=\frac{t}{e^t-1}$ (FFT-able).
+		$B[0,\ldots] = [1, -\frac{1}{2}, \frac{1}{6}, 0, -\frac{1}{30}, 0, \frac{1}{42}, \ldots]$
+
+		Sums of powers:
+		\small
+		\[ \sum_{i=1}^n n^m = \frac{1}{m+1} \sum_{k=0}^m \binom{m+1}{k} B_k (n+1)^{m+1-k} \]
```
simonlindholm: `B_k` reads like a function here, but I guess it's really a number? If so maybe add a `\cdot` after it

content/combinatorial/chapter.tex
```diff
@@ -65,6 +80,11 @@ \section{General purpose numbers}
 		$1, 1, 2, 5, 15, 52, 203, 877, 4140, 21147, \dots$. For $p$ prime,
 		\[ B(p^m+n)\equiv mB(n)+B(n+1) \pmod{p} \]
 
+	\subsection{Labeled unrooted trees}
+		\# on $n$ vertices: $n^{n-2}$ \\
+		\# on $k$ existing trees of size $n_i$: $n_1n_2\cdots n_k n^{k-2}$ \\
+		\# with degrees $d_i$: $(n-2)! / ((d_1-1)! \cdots (d_n-1)!)$
```
simonlindholm: I don't get this part. If every node has degree 1 this formula says the number of trees should be `(n-2)!`, but it's actually 0?

Chillee: hmm, yeah, it doesn't seem to be quite correct, looking here: https://cp-algorithms.com/graph/pruefer_code.html#toc-tgt-7

Seems like the correct formula is , where `s_i` is the size of the i-th connected component, where we have `k` components to begin with.

simonlindholm: Apparently I think I just missed the silent assumption that the sum of the degrees must be 2k - 2 for the formula to hold.

PR #150, opened 2019-11-23
Update documentation of ConvexHull.h

aleloi:
It says in the pdf that this function returns the *indices* of the convex hull. That's not correct any longer as it returns the actual points (see the return type and line 32).


simonlindholm:
Thanks!

PR #153, opened 2019-12-28
Directed MST reconstruction

### LINKED ISSUE #120: Directed MST edge reconstruction
The arborescence implementation added in #110 does not support reconstructing the edges. Might be nice to have if it's not too long.

simonlindholm:
Doesn't add too many lines, nor does it affect perf at all. Also adds a supporting DSU with rollback. The most controversial part of this is removing the standard DSU from the pdf.

Closes #120.


Chillee:
IMO, just mark the lines that don't need to be included if you don't care about rollback (maybe with a comment) and it's perfectly fine.

simonlindholm:
Updated with a short note in the UnionFindRollback description.

PR #154, opened 2019-12-29
Add MoQueries

simonlindholm:
Still remaining: something about queries on trees (comment or code, not sure).


Chillee:
I'm not sure this description is enough for someone to perform mo on trees.

simonlindholm:
> I'm not sure this description is enough for someone to perform mo on trees.

Added some code for it. Turned out to be kinda tricky to implement, making this potentially worth including in the pdf by default.

PR #155, opened 2020-01-12
Replaced HLD with much better version

Chillee:
Pretty neat implementation.

Integrates this: https://codeforces.com/blog/entry/53170

Will write fuzz test in morning - have tested on small examples only.


content/graph/HLD.h
```diff
@@ -1,112 +1,60 @@
...
-					Chain& c = C[n1.chain];
-					f(ans, n1.pos ? c.tree.query(n1.pos, sz(c.nodes))
-					              : c.tree.s[1]);
-					i1 = c.par;
-				}
-			}
+	template <class B> void process(int u, int v, B op) {
+		for (; rt[u] != rt[v]; v = par[rt[v]]) {
+			if (depth[rt[u]] > depth[rt[v]]) swap(u, v);
+			op(pos[rt[v]], pos[v] + 1);
```
simonlindholm: Not super happy that we assume commutativity here, but I guess fixing that takes up a bunch of space (and is also doable during contest)

content/graph/HLD.h
```diff
@@ -1,112 +1,60 @@
 /**
- * Author: Johan Sannemo
- * Date: 2015-02-10
+ * Author: Benjamin Qi, Oleksandr Kulkov, chilli
+ * Date: 2020-01-12
  * License: CC0
```
simonlindholm: reminder to check license

content/graph/HLD.h
```diff
@@ -1,112 +1,60 @@
...
-
-	HLD(vector<vpi>& g) : V(sz(g)) {
-		dfs(0, -1, g, 0);
-		trav(c, C) {
-			c.tree = {sz(c.nodes), 0};
-			for (int ni : c.nodes)
-				c.tree.update(V[ni].pos, V[ni].val);
+#include "../data-structures/LazySegmentTree.h"
+
+template <bool USE_EDGES> struct HLD {
```
simonlindholm: what does USE_EDGES mean? that nodes have the weight of the edge to their parent?

content/graph/HLD.h
Chillee: Already asked.

content/graph/HLD.h
Chillee: It's whether the values are stored in the edges or in the nodes.

ie: if you have a path (0-1-2) and you modify it, are you modifying 3 values or 2?

The original name was `VALS_IN_EDGES` which I thought was a bit too long - I suppose it might be a more descriptive name.

Regardless, I should definitely add it to the description.

simonlindholm: I think the longer name is probably better (or a shorter + comment in the description), but I still don't totally get it: where concretely are the edge values stored if it's set? I'm speculating each node contains the weight of the edge to its parent.

content/graph/HLD.h
Chillee: Resolved in another thread - supporting non-commutative functions requires substantially more code and is fairly rare.

Chillee:
Stress-tested against old HLD as well as brute force implementation. Only functionality I haven't tested is `VALS_EDGES` (all the tests assume it's false).

I also modified the lazy segtree a bit to abstract the op out. This is primarily because the HLD also needs to use this operation, and I think it's pretty ugly to have an invisible dependency.

PR #158, opened 2020-02-19
Remove distance from LCA

### LINKED ISSUE #146: Remove distance from LCA
It's seldom useful, easy to add manually and costs lines.

simonlindholm:
Fixes #146. Removes 8 lines of code and 4 lines of description, and makes it more obvious how to use the LCA. At the same time, if you do need this, it's easy to re-introduce manually.

### DIFF
--- a/content/graph/LCA.h
+++ b/content/graph/LCA.h
@@ -27,10 +27,10 @@ struct LCA {
 		}
 	}
 
-	int query(int a, int b) {
+	int lca(int a, int b) {
 		if (a == b) return a;
 		tie(a, b) = minmax(time[a], time[b]);
 		return path[rmq.query(a, b)];
 	}
-	// dist(a,b) { return dist[a] + dist[b] - 2*dist[lca(a,b)]; }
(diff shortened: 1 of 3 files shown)


content/graph/LCA.h
```diff
@@ -3,35 +3,27 @@
...
- * Status: Somewhat tested
+ * Status: stress-tested
  * Description: Data structure for computing lowest common ancestors in a tree
  * (with 0 as root). C should be an adjacency list of the tree, either directed
- * or undirected. Can also find the distance between two nodes.
- * Usage:
- *  LCA lca(undirGraph);
- *  lca.query(firstNode, secondNode);
- *  lca.distance(firstNode, secondNode);
+ * or undirected.
```
Chillee: I think it's worth adding a comment about formula for distance between 2 nodes. Would help beginner teams.

simonlindholm: Added as a commented-out function. Not 100% sure I agree with it being useful, but let's go with it, it's just one line after all.

PR #159, opened 2020-03-05
Updated Pollard Rho (also added a details commands for implementation details)

Chillee:
Accumulates GCD calls, resulting in about a 6-10x runtime improvement.

I find that when teams use KACTL, they'll often replace an implementation with their own implementation if they don't understand why something was done. Implementation details is an opportunity to explain the tradeoffs that were made without taking up valuable document space. Much of this discussion has been done in chat or PR's previously - I think it's worth adding them in the code itself.

I chose the tag "Details".

### DIFF
--- a/content/number-theory/Factor.h
+++ b/content/number-theory/Factor.h
@@ -1,11 +1,16 @@
 /**
- * Author: chilli, SJTU
- * Date: 2019-04-24
+ * Author: chilli, SJTU, pajenegod
+ * Date: 2020-03-04
  * License: CC0
- * Source: https://github.com/FTRobbin/Dreadnought-Standard-Code-Library
+ * Source: own
  * Description: Pollard-rho randomized factorization algorithm. Returns prime
  * factors of a number, in arbitrary order (e.g. 2299 -> \{11, 19, 11\}).
  * Time: $O(n^{1/4})$ gcd calls, less for numbers with small factors.
+ * Details: This implementation uses the improvement described here: https://en.wikipedia.org/wiki/Pollard%27s_rho_algorithm#Variants, where one can accumulate gcd calls by some factor (40 chosen here through exhaustive testing). This improves performance by approximately 6-10x depending on the inputs and speed of gcd. Benchmark found here: https://ideone.com/nGGD9T
+ * GCD can improved by a factor of 1.75x using Binary GCD (https://lemire.me/blog/2013/12/26/fastest-way-to-compute-the-greatest-common-divisor/). However, with the gcd accumulation the bottleneck moves from the gcd calls to the mod_mul. As GCD only constitutes ~12% of runtime, speeding it up doesn't matter so much.
+ * This code can probably be sped up by using a faster mod mul - potentially montgomery reduction on 128 bit integers.
+ * Alternatively, one can use a quadratic sieve for an asymptotic improvement, which starts being factor in practice around 1e13.
+ * Brent's cycle finding algorithm was tested. It reduces f calls by a small amount in exchange for a significant increase in gcd calls.
  */
 #pragma once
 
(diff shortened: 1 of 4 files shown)


content/number-theory/Factor.h
```diff
@@ -1,26 +1,51 @@
...
 ull pollard(ull n) {
-	auto f = [n](ull x) { return (mod_mul(x, x, n) + 1) % n; };
+	auto f = [n](ull x) { return mod_mul(x, x, n) + 1; };
 	if (!(n & 1)) return 2;
-	for (ull i = 2;; i++) {
-		ull x = i, y = f(x), p;
-		while ((p = __gcd(n + y - x, n)) == 1)
-			x = f(x), y = f(f(y));
-		if (p != n) return p;
+	ull x = 0, y = 0, tim = 0, prd = 1, i = 1, tmp;
```
simonlindholm: `i` starting at `1` instead of `0` is kinda subtle, I need to think about this...

content/number-theory/Factor.h
```diff
@@ -1,26 +1,51 @@
...
+ * to the mod_mul. As GCD only constitutes ~12% of runtime, speeding it up
+ * doesn't matter so much.
+ *
+ * This code can probably be sped up by using a faster mod mul - potentially
+ * montgomery reduction on 128 bit integers.
+ * Alternatively, one can use a quadratic sieve for an asymptotic improvement,
+ * which starts being factor in practice around 1e13.
+ *
+ * Brent's cycle finding algorithm was tested, but doesn't reduce mod_mul calls
+ * significantly.
```
simonlindholm: add `Status: stress-tested`

content/number-theory/Factor.h
```diff
@@ -1,26 +1,51 @@
...
+	auto f = [n](ull x) { return mod_mul(x, x, n) + 1; };
 	if (!(n & 1)) return 2;
-	for (ull i = 2;; i++) {
-		ull x = i, y = f(x), p;
-		while ((p = __gcd(n + y - x, n)) == 1)
-			x = f(x), y = f(f(y));
-		if (p != n) return p;
+	ull x = 0, y = 0, tim = 0, prd = 1, i = 1, tmp;
+	for (; prd; y = f(f(y)), x = f(x)) {
+		if (x == y || ++tim == 40) {
```
simonlindholm: could do `++tim % 32 == 0` to save the next line if it doesn't affect perf. In that case the if statements could even be combined, and the while loop moved to after the if

content/number-theory/chapter.tex
```diff
@@ -48,3 +48,12 @@ \section{Estimates}
...
 	The number of divisors of $n$ is at most around 100 for $n < 5e4$, 500 for $n < 1e7$, 2000 for $n < 1e10$, 200\,000 for $n < 1e19$.
+
+\section{Mobius Function}
+\[
+	\mu(n) = \begin{cases} 0 & n \textrm{ is not square free}\\ 1 & n \textrm{ has even number of prime factors}\\ -1 & n \textrm{ has odd number of prime factors}\\\end{cases}
+\]
+  Mobius Inversion: If $g(n) = \sum_{d|n} f(d)$, then $f(n) = \sum_{d|n} \mu(d)g(n/d)$.
+  \[
+	  \sum_{d | n} \mu(d) = \begin{cases} 1 & \text{if} n = 1 \\ 0 \text{if} n > 1 \end{cases}
+  \]
```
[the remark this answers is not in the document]
Chillee: Oh whups, didn't realize I'd added this in. I'll add in some other formulas too, including `\sum_{d|n} \mu(d) = [n = 1]`.

content/number-theory/Factor.h
```diff
@@ -1,26 +1,51 @@
...
+ * significantly.
  */
 #pragma once
 
 #include "ModMulLL.h"
 #include "MillerRabin.h"
 
 ull pollard(ull n) {
-	auto f = [n](ull x) { return (mod_mul(x, x, n) + 1) % n; };
+	auto f = [n](ull x) { return mod_mul(x, x, n) + 1; };
```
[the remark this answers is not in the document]
Chillee: Yeah it makes a pretty significant difference. Calls to `f` are the primary bottleneck now, so removing an extra modulo operation results in about 1.5x increase in speed.

content/number-theory/Factor.h
Chillee: It doesn't really matter I think. Arguably, `i` should be randomized. In the current code, it starts at 2.

content/number-theory/Factor.h
Chillee: Still using 40 (doesn't seem to affect perf). For every mod we do, we call `mod_mul` at least 4 times.

content/number-theory/chapter.tex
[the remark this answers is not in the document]
Chillee: Actually, I'll leave this for a later PR

Chillee:
Golfed it a bit more thanks to pajenegod on discord.

Chillee:
Most of the golfing is thanks to pajenegod - thanks a lot to him :)

PR #161, opened 2020-03-23
Fixed inconsistencies between pushrelabel and dinic

Chillee:
Maybe I should have just pushed this.


simonlindholm:
Yep, looks pretty clear-cut to me.

PR #164, opened 2020-04-03
Added Gomory Hu Tree

### LINKED ISSUE #50: Gomory-Hu tree
Not sure how useful it is, but it seems like the code is rather short, and it would be hard to derive by hand.

Chillee:
Fixes #50 .

We can use any flow algorithm as long as it allows us to call an equivalent function to `leftOfMinCut`. I decided to make it a function instead of a struct because it's relatively simple.

We also don't really care about passing the output to `binaryLifting`, as the runtime will be dominated by the flow.


stress-tests/utilities/utils.h
```diff
@@ -0,0 +1,15 @@
+#pragma once
+
+template <class F> struct y_combinator {
+    F f; // the lambda will be stored here
+
+    // a forwarding operator():
+    template <class... Args> decltype(auto) operator()(Args &&... args) const {
+        // we pass ourselves to f, then the arguments.
+        // [edit: Barry] pass in std::ref(*this) instead of *this
```
simonlindholm, line 9: silly comment. :) link directly to https://stackoverflow.com/questions/2067988/recursive-lambda-functions-in-c11 next to the struct instead?

stress-tests/graph/GomoryHu.cpp
```diff
@@ -0,0 +1,46 @@
+#include "../utilities/template.h"
+#include "../utilities/utils.h"
+
+#include "../../content/graph/GomoryHu.h"
+#include "../../content/graph/Dinic.h"
+
+
+void test(int N, int mxFlow, int iters) {
+    for (int it = 0; it < iters; it++) {
```
simonlindholm, line 9: set `n = rand() % N` or `n = rand() % N + 1` here and use that n instead to improve test coverage

stress-tests/graph/GomoryHu.cpp
```diff
@@ -0,0 +1,46 @@
...
+                    dfs(start, i[0], cur, min(mn, i[1]));
+            }
+        });
+        dfs(0, 0, -1, 1e9);
+    }
+}
+signed main() {
+    test(25, 5, 105);
+    test(100, 1000, 5);
+    test(100, 1, 50);
```
simonlindholm, line 44: needs more tiny tests. `test(5, 5, 10000);`?

content/graph/GomoryHu.h
```diff
@@ -0,0 +1,27 @@
...
+ * Description: Given a list of edges representing an undirected flow graph, returns edges of the Gomory-Hu tree. The max flow between any pair of
+ * vertices is given by minimum edge weight along the Gomory-Hu tree path.
+ * Time: $O(V)$ Flow Computations
+ * Status: Tested on CERC 2015 J, stress-tested
+ */
+#pragma once
+
+#include "PushRelabel.h"
+
+typedef array<ll, 3> edge;
```
simonlindholm, line 15: I wonder if PushRelabel's Edge struct should be made internal so that this can be named `Edge`. But, eh, having the same name may still just be confusing.

Chillee: having `edge` and `Edge` is going to be confusing anyways.I just moved it in.

PR #165, opened 2020-04-04
Updated Hungarian

Chillee:
I'll stress test it against MCMF tomorrow.

### DIFF
--- a/content/graph/WeightedMatching.h
+++ b/content/graph/WeightedMatching.h
@@ -1,18 +1,21 @@
 /**
  * Author: Benjamin Qi, Chilli
- * Date: 4/04/20
+ * Date: 2020-04-04
  * License: CC0
  * Source: https://github.com/bqi343/USACO/blob/master/Implementations/content/graphs%20(12)/Matching/Hungarian.h
- * Description: Minimum cost bipartite matching. Takes in cost[N][M], where N
- * <= M, and cost[i][j] = cost for i-th node on left to j-th node on right.
- * Returns flow and match, where match[i] = node assigned to i-th left node.
- * Negate costs for max cost.
+ * Description: Given array of (possibly negative) costs to complete $N$ jobs
+ * w/ $M$ workers $(N \le M)$, finds min cost to complete all jobs s.t. each
+ * worker is assigned to at most one job. Takes cost[N][M], where cost[i][j] =
+ * cost for i-th job to be completed by j-th worker and returns (min cost,
+ * match), where match[i] = worker assigned to i-th job. Negate costs for max
+ * cost.
  * Time: O(N^2M)
  * Status: Tested on kattis:cordonbleu
  */
 #pragma once
 
 pair<int, vi> hungarian(const vector<vi> &a) {
+	if (a.empty()) return {0, {}};
 	int n = sz(a) + 1, m = sz(a[0]) + 1;
 	vi u(n), v(m), p(m), ans(n - 1);
 	rep(i,1,n) {
--- a/content/graph/WeightedMatching.h
+++ b/content/graph/WeightedMatching.h
@@ -10,7 +10,7 @@
  * match), where match[i] = worker assigned to i-th job. Negate costs for max
  * cost.
  * Time: O(N^2M)
- * Status: Tested on kattis:cordonbleu
+ * Status: Tested on kattis:cordonbleu, stress-tested
  */
 #pragma once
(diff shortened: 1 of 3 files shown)


content/graph/WeightedMatching.h
```diff
@@ -1,89 +1,44 @@
 /**
- * Author: Stanford
- * Date: Unknown
- * Source: Stanford Notebook
- * Description: Min cost bipartite matching. Negate costs for max cost.
- * Time: O(N^3)
- * Status: tested during ICPC 2015
+ * Author: Benjamin Qi, Chilli
+ * Date: 4/04/20
```
simonlindholm, line 3: use a sane date format

content/graph/WeightedMatching.h
```diff
@@ -1,89 +1,44 @@
...
+ * License: CC0
+ * Source: https://github.com/bqi343/USACO/blob/master/Implementations/content/graphs%20(12)/Matching/Hungarian.h
+ * Description: Minimum cost bipartite matching. Takes in cost[N][M], where N
+ * <= M, and cost[i][j] = cost for i-th node on left to j-th node on right.
```
simonlindholm, line 7: say something about every node on the left choosing one on the right such that no two choose the same node. I.e., make this intelligible without knowing what it does

content/graph/WeightedMatching.h
```diff
@@ -1,89 +1,44 @@
...
+ * Returns flow and match, where match[i] = node assigned to i-th left node.
```
simonlindholm, line 8: flow is a confusing word

content/graph/WeightedMatching.h
```diff
@@ -1,89 +1,44 @@
...
-			L[R[j]] = j;
-			j = d;
-		}
-		R[j] = s;
-		L[s] = j;
 	}
-	auto value = vd(1)[0];
-	rep(i,0,n) value += cost[i][L[i]];
-	return value;
+	rep(j,1,m) if (p[j]) ans[p[j] - 1] = j - 1;
```
simonlindholm, line 42: maybe using `n`/`m` as the dummy nodes could make this shorter

content/graph/WeightedMatching.h
```diff
@@ -1,89 +1,44 @@
...
-			dist[k] = cost[s][k] - u[s] - v[k];
-
-		int j = 0;
-		for (;;) { /// find closest
-			j = -1;
-			rep(k,0,n){
-				if (seen[k]) continue;
-				if (j == -1 || dist[k] < dist[j]) j = k;
+pair<int, vi> hungarian(const vector<vi> &a) {
+	int n = sz(a) + 1, m = sz(a[0]) + 1;
```
simonlindholm, line 16: return early if a.empty()

content/graph/WeightedMatching.h
Chillee: it was late :'(

content/graph/WeightedMatching.h
Chillee: I think it's a bit uglier doing this.
```cpp
pair<int, vi> hungarian(const vector<vi> &a) {
    if (a.empty()) return {0, {}};
    int n = sz(a), m = sz(a[0]) + 1;
    vi u(n), v(m), p(m, -1), ans(n - 1);
    rep(i, 0, n) {
        p[m - 1] = i;
        int j0 = m - 1; // add "dummy" worker 0
        vi dist(m, INT_MAX), pre(m, -1);
        vector<bool> done(m + 1);
        do { // dijkstra
            done[j0] = true;
            int i0 = p[j0], j1, delta = INT_MAX;
            rep(j, 0, m - 1) if (!done[j]) {
                auto cur = a[i0][j] - u[i0] - v[j];
                if (cur < dist[j]) dist[j] = cur, pre[j] = j0;
                if (dist[j] < delta) delta = dist[j], j1 = j;
            }
            rep(j, 0, m) {
                if (done[j]) u[p[j]] += delta, v[j] -= delta;
                else dist[j] -= delta;
            }
            j0 = j1;
        } while (p[j0] != -1);
        while (j0 != m - 1) { // update alternating path
            int j1 = pre[j0];
            p[j0] = p[j1], j0 = j1;
        }
    }
    rep(j, 0, m - 1) if (p[j]) ans[p[j]] = j;
    return {-v[m - 1], ans}; // min cost
}
```

PR #166, opened 2020-04-13
Replaced sieve with seg sieve

Chillee:
I've asked Jakob Kogler about the license, and he's fine with CC0.

Can sieve up to `1e9` in ~1.3 seconds (including on CF), while current KACTL sieve takes ~5.5 seconds.


content/number-theory/eratosthenes.h
```diff
@@ -1,22 +1,39 @@
...
+	}
+	for (int l = 1; l <= h; l += S) {
+		array<bool, S> block{};
+		trav(i, cp) {
+			int idx = i[1];
+			for (; idx < S; idx += i[0]) block[idx] = true;
+			i[1] = idx - S;
+		}
+		rep(i,0,min(S, h - l))
+			if (!block[i]) pr.push_back((l + i) * 2 + 1);
```
simonlindholm, line 36: I guess adding an isprime char array and setting isprime[...] = 1 here wouldn't lose too much performance, though it would be bad for memory usage...

content/number-theory/eratosthenes.h
```diff
@@ -1,22 +1,39 @@
...
-		for (int j = i*i; j < lim; j += i*2) isprime[j] = 0;
-	vi pr;
-	rep(i,2,lim) if (isprime[i]) pr.push_back(i);
+const int LIM = 1e6;
+vi eratosthenes() {
+	const int S = round(sqrt(LIM)), h = (LIM - 1) / 2;
+	vi pr({2}), sieve(S + 1);
+	vector<array<int, 2>> cp;
+	for (int i = 3; i < S; i += 2) {
+		if (sieve[i]) continue;
```
simonlindholm, line 24:
```
for (...) if (...) {
}
```

content/number-theory/eratosthenes.h
```diff
@@ -1,22 +1,39 @@
...
-bitset<MAX_PR> isprime;
-vi eratosthenesSieve(int lim) {
-	isprime.set(); isprime[0] = isprime[1] = 0;
-	for (int i = 4; i < lim; i += 2) isprime[i] = 0;
-	for (int i = 3; i*i < lim; i += 2) if (isprime[i])
-		for (int j = i*i; j < lim; j += i*2) isprime[j] = 0;
-	vi pr;
-	rep(i,2,lim) if (isprime[i]) pr.push_back(i);
+const int LIM = 1e6;
+vi eratosthenes() {
```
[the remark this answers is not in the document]
Chillee: It's not a huge amount of performance but it's like ~10% or so.

content/number-theory/eratosthenes.h
Chillee: Pretty much negligible performance drop AFAICT.

content/number-theory/eratosthenes.h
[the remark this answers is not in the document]
Chillee: What's the reasoning behind wanting LIM as a parameter?

content/number-theory/eratosthenes.h
```diff
@@ -1,22 +1,39 @@
...
+	const int S = round(sqrt(LIM)), h = (LIM - 1) / 2;
```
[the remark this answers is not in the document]
Chillee: This doesn't throw `-Wconversion warnings` for me.

content/number-theory/Eratosthenes.h
```diff
@@ -0,0 +1,36 @@
...
+ *
+ * Benchmark can be found here: https://ideone.com/e7TbX4
+ */
+#pragma once
+
+const int LIM = 1e6;
+bitset<LIM> isPrime;
+vi eratosthenes() {
+	const int S = round(sqrt(LIM)), R = LIM / 2;
+	vi pr({2}), sieve(S + 1); pr.reserve(LIM / (int)log(LIM));
```
jakobkogler, line 21: If you want to reserve the memory for `pr`, you probably should use a bigger value. The actual number of primes is a little bigger than `LIM / log(LIM)`, e.g.  ~6% for `LIM = 10^8`. So this still requires the vector to resize once, ending with almost twice the needed memory. For reference: https://en.wikipedia.org/wiki/Prime-counting_function#Table_of_%CF%80(x),_x_/_ln_x,_and_li(x)
Also, shouldn't the conversion to `int` happen after the division?

I suggest something like `(int)(LIM / log(LIM) * 1.1)` which fits for all `LIM >= 10^5`. Or `(int)(LIM / log(LIM) * 1.07` for all `LIM >= 10^7)`. (For smaller `LIM` the vector of primes is small anyhow, and it doesn't matter how much memory we use.

content/number-theory/Eratosthenes.h
```diff
@@ -0,0 +1,36 @@
...
+	vector<array<int, 2>> cp;
+	for (int i = 3; i <= S; i += 2) if (!sieve[i]) {
+		cp.push_back({i, i * i / 2});
+		for (int j = i * i; j <= S; j += 2 * i) sieve[j] = 1;
+	}
+	for (int L = 1; L <= R; L += S) {
+		array<bool, S> block{};
+		for (auto &[p, idx] : cp)
+			for (int i=idx; i < S+L; idx = (i+=p)) block[i-L] = 1;
```
jakobkogler, line 30: This line is much shorter as `for (; idx < S + L; idx += p) block[idx-L] = 1;`?
Be careful. I haven't tested it.

Chillee: Unfortunately, this was done on purpose :'(  Since `idx` is a reference into the `cp` array, accessing it is slower than accessing a local variable.

Ideally the compiler would be able to hoist it out of the loop, but it seems that the mixed load/store instructions confuse the compiler. The way it's written, there's only a single load to `idx` and then repeated stores, which the compiler seems to optimize fine.

It causes a `~200ms` difference in our testing - I'll write a note about it.

content/number-theory/Eratosthenes.h
Chillee: Good point - I was just being lazy. Considering that this is needed to squeeze the memory to `200MB` though, I think this is a good idea.

content/number-theory/FastEratosthenes.h
```diff
@@ -0,0 +1,39 @@
...
+ * Se https://github.com/kth-competitive-programming/kactl/pull/166#discussion_r408354338
+ */
+#pragma once
+
+const int LIM = 1e6;
+bitset<LIM> isPrime;
+vi eratosthenes() {
+	const int S = round(sqrt(LIM)), R = LIM / 2;
+	vi pr = {2}, sieve(S+1); pr.reserve(int(LIM/log(LIM)*1.1));
+	vector<array<int, 2>> cp;
```
simonlindholm, line 25: can be shortened to `vector<pii>` without any further changes. alternatively, could use [0]/[1] below to get rid of the C++17 for loop, either way.

PR #169, opened 2020-04-14
Added mobius inversion

Chillee:
I'm not sure what you meant by this: https://github.com/kth-competitive-programming/kactl/pull/159#discussion_r388126994

I'm not aware of such a version - are you thinking of what he reduced the problem to here? https://codeforces.com/blog/entry/53925

Looks like this: 

image

### DIFF
--- a/content/number-theory/chapter.tex
+++ b/content/number-theory/chapter.tex
@@ -48,3 +48,14 @@
 	$\sum_{d|n} d = O(n \log \log n)$.
 
 	The number of divisors of $n$ is at most around 100 for $n < 5e4$, 500 for $n < 1e7$, 2000 for $n < 1e10$, 200\,000 for $n < 1e19$.
+
+\section{Mobius Function}
+\[
+	\mu(n) = \begin{cases} 0 & n \textrm{ is not square free}\\ 1 & n \textrm{ has even number of prime factors}\\ -1 & n \textrm{ has odd number of prime factors}\\\end{cases}
+\]
+  Mobius Inversion:
+  \[ g(n) = \sum_{d|n} f(d) \Leftrightarrow f(n) = \sum_{d|n} \mu(d)g(n/d) \]
+
+  Other useful formulas/forms:
+  \[ \sum_{d | n} \mu(d) = [ n = 1] \]
+  \[ g(n) = \sum_{n|d} f(d) \Leftrightarrow f(n) = \sum_{n|d} g(d)\mu(d/n)\]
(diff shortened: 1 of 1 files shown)


simonlindholm:
> I'm not sure what you meant by this: #159 (comment)

https://en.wikipedia.org/wiki/M%C3%B6bius_inversion_formula#Generalizations

content/number-theory/chapter.tex
```diff
@@ -48,3 +48,14 @@ \section{Estimates}
...
+\section{Mobius Function}
+\[
+	\mu(n) = \begin{cases} 0 & n \textrm{ is not square free}\\ 1 & n \textrm{ has even number of prime factors}\\ -1 & n \textrm{ has odd number of prime factors}\\\end{cases}
+\]
+  Mobius Inversion:
+  \[ g(n) = \sum_{d|n} f(d) \Leftrightarrow f(n) = \sum_{d|n} \mu(d)g(n/d) \]
+
+  Other useful formulas/forms:
+  \[ \sum_{d | n} \mu(d) = [ n = 1] \]
+  \[ g(n) = \sum_{n|d} f(d) \Leftrightarrow f(n) = \sum_{n|d} g(d)\mu(d/n)\]
```
simonlindholm, line 61: `g(n) = \sum_{n|d} f(d)` seems to be a sum over infinitely many d?

Chillee: https://math.stackexchange.com/questions/1965342/does-the-mobius-inversion-theorem-hold-when-sum-function-is-over-multiples-inste

and

https://codeforces.com/blog/entry/59962?#comment-475400

It's possible if `f(d) = 0` for some high enough d.

Chillee:
Added the floor formula. Also switched to `$` instead of `\[` since it took up a lot of space. I don't know if there's a better way to do it.

Current:

image

Originally:

image

content/number-theory/chapter.tex
simonlindholm: Oh, I see. I wonder if it's worth a comment, but not obvious what would fit. "(if convergent)" maybe

content/number-theory/chapter.tex
```diff
@@ -48,3 +48,17 @@ \section{Estimates}
...
+
+\section{Mobius Function}
+\[
+	\mu(n) = \begin{cases} 0 & n \textrm{ is not square free}\\ 1 & n \textrm{ has even number of prime factors}\\ -1 & n \textrm{ has odd number of prime factors}\\\end{cases}
+\]
+  Mobius Inversion:
+  \[ g(n) = \sum_{d|n} f(d) \Leftrightarrow f(n) = \sum_{d|n} \mu(d)g(n/d) \]
+  Other useful formulas/forms:
+
+  $ \sum_{d | n} \mu(d) = [ n = 1] $ (very useful)
```
simonlindholm, line 60: remove this comment since it line-wraps, I guess

Issue #116, opened 2019-05-11
Candidates for removal


Chillee:
Since with the new additions in PR we're running fairly close to the page limit, I thought it'd be a good idea to keep a running list of what might be removable.

SubMatrix.h (13 lines)- Pretty trivial to re-implement from scratch.
Matrix.h (26 line)- Also pretty trivial to re-implement. It isn't used elsewhere either.
FenwickTree.h (22 lines) - I personally find 1d fenwick trees to be nearly completely useless - especially since we have the bottom-up segtree that runs nearly as fast as the fenwick tree.
Bellman Ford (12 line)/Floyd Warshall (12 lines) - Also pretty trivial

simonlindholm:
> I personally find 1d fenwick trees to be nearly completely useless - especially since we have the bottom-up segtree that runs nearly as fast as the fenwick tree.

How large is the perf difference? Also, we should probably add some code to segtree in this case to support lower_bound.

Also, perhaps:
* [ ] line-hull intersection
* [ ] global min cut
* [ ] TopoSort
* [ ] binomialModPrime should be just a description
* [ ] HillClimbing
* [ ] Polynomial

On the topic of shorting we could also decrease the font size for surrounding .tex code -- the algorithms are currently of rather smaller size than the math, while being used far more frequently.

General purpose numbers and Probability distributions don't really pull their weight. They could be made shorter by using tables, I guess.

Chillee:
Well, nearly as fast is a bit of an exaggeration - it's a 55% difference.

simonlindholm:
> On the topic of shorting we could also decrease the font size for surrounding .tex code -- the algorithms are currently of rather smaller size than the math, while being used far more frequently.

Done in 3ad3b65505e1d1707a66255575757c2692149abd. (Won two whole columns; we're currently at 23.666 pages)

Chillee:
Let's include some of the stuff we're not currently including then :)

Currently these are the files excluded:
```
./content/numerical/MatrixInverse-mod.h
./content/various/FastInput.h
./content/tex/preprocessor.py
./content/graph/Dinic.h
./content/graph/GomoryHu.h
./content/strings/Hashing-codeforces.h
./content/geometry/CirclePolygonIntersection.h
./content/geometry/PolygonUnion.h
./content/geometry/ManhattanMST.h
./content/geometry/LineProjectionReflection.h
./content/geometry/DelaunayTriangulation.h
./content/geometry/CircleLine.h
```

[...]

PR #170, opened 2020-04-14
Moved binomialmodprime to description

Chillee:
In accordance with https://github.com/kth-competitive-programming/kactl/issues/116#issuecomment-491495055, I moved `binomialmodprime` to description.

I also don't see why we have multinomial? If you have the equation it seems fairly trivial to implement, unless I'm missing something.


simonlindholm:
> If you have the equation it seems fairly trivial to implement, unless I'm missing something.

Overflows makes it a bit tricky. I'd prefer to keep it, it's short anyway.

PR #171, opened 2020-04-14
[Speculative] Tried replacing phiFunction with linear sieve that can calculate mobius as well.

Chillee:
Not sure I like it. There might be other better/more general ways of calculating multiplicative functions. On the other hand, if we're including mobius inversion I think it makes sense to have a way of calculating the mobius function.

### DIFF
--- a/content/number-theory/phiFunction.h
+++ b/content/number-theory/phiFunction.h
@@ -1,27 +1,29 @@
 /**
- * Author: Håkan Terelius
- * Date: 2009-09-25
+ * Author:
+ * Date:
  * License: CC0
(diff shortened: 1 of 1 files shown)


simonlindholm:
Seems kinda non-obvious how to change. I think a useful primitive might be a sieve that computes the smallest (or largest) prime that divides every number. Then it would be easy to compute phi:
```cpp
rep(a,2,n) {
  int p = smallestPrimeDivisor[a], b = a / p;
  phi[a] = phi[b] * (p - (smallestPrimeDivisor[b] != p));
}
```
This may be too easy to code by hand to be worth including though.

PR #174, opened 2020-05-02
Make modmul faster for constant moduli

simonlindholm:
Changes `ull(ld(a) * ld(b) / ld(M))` -> `ull(1.L / M * a * b);` to win a small bit of performance when `M` is constant over consecutive modmuls, and to shave off a few characters. Also adds a 4-page proof for the new version being correct for up to `M < 0.394 * 2^64`. (Somewhat excessive I guess, but it was fun.)

### DIFF
--- a/content/number-theory/MillerRabin.h
+++ b/content/number-theory/MillerRabin.h
@@ -4,7 +4,7 @@
  * License: CC0
  * Source: Wikipedia, https://miller-rabin.appspot.com/
  * Description: Deterministic Miller-Rabin primality test.
- * Guaranteed to work for numbers up to $2^{63}$; for larger numbers, extend A randomly.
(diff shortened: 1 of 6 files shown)


simonlindholm:
Rendered pdf: https://www.overleaf.com/read/yymxmzsfcbbj

khanhhungvu1508:
Hello you. I have a question to you. I don't know why we have that during the proof when c > 2^62?
image

simonlindholm:
r(x) is very very close to x multiplicatively (unless we hit denormals which we aren't close to here), so if we throw in some r()'s in 1/c * a * b we'll only blow it up by a small factor; 1.001 is extremely generous. And a * b / c < c, so even if we scale that up by 1.001, we'll be less than 2^63 by a factor of 1 / (0.394 * 2 * 1.001).

PR #175, opened 2020-06-03
More general ModLog

bjorn-martinsson:
This PR adds two new features to ModLog(a,b,m).

1. The return value is now smallest x > 0, instead of smallest x >= 0. This allows for being able to calculate the order of an element a by doing ModLog(a,1,m).
2. It is no longer required for a and m to be coprime.

The algorithm is a modified version of the standard baby-step giant-step method. The modification allows for even non-coprime input using an original idea of mine.

Standard baby-step giant-step is based around finding (i, j) such that a^(i * n) = b * a^j (mod m), where n > sqrt(m) and 0<i,j<=n. If a and m are coprime, then a^j has a modular inverse, which means a^(i * n - j) = b (mod m). 

The modification (allowing for non-coprime a and m) is based around the following idea.

Assume p^x is a prime power divisor of m.
Then we have 3 cases
1. b is divisible by p^x
2. b is divisible only by some p^y, 0<y<x
3. b is not divisible by p

The important thing to note is that in case 2, the discrete log (if it exists) cannot be > sqrt(m), (technically it cannot be >= log2(m)). So once you've checked all exponents of a that are <= sqrt(m) you cannot have case 2. Case 2 is the only tricky case. 

So the modification involves checking all exponents of a that are <= n, and then handling the non-tricky cases by a simple gcd(a^n, m) = gcd(b, m) check.

### DIFF
--- a/content/number-theory/ModLog.h
+++ b/content/number-theory/ModLog.h
@@ -1,21 +1,22 @@
 /**
- * Author: Andrei Navumenka, chilli
- * Date: 2019-11-08
- * License: Unlicense
- * Source: https://github.com/indy256/codelibrary/blob/master/cpp/numbertheory/discrete_log.cpp
- * Description: Returns the smallest $x \ge 0$ s.t. $a^x = b \pmod m$. a and m must be coprime.
+ * Author: Bjorn Martinsson
+ * Date: 2020-06-03
+ * License: CC0
+ * Source: own work
+ * Description: Returns the smallest $x > 0$ s.t. $a^x = b \pmod m$.
  * Time: $O(\sqrt m)$
- * Status: tested for all 0 <= a,x,m < 200.
+ * Status: tested for all 0 <= a,x < 500 and 0 < m < 500.
  */
 #pragma once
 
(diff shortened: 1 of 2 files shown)


content/number-theory/ModLog.h
```diff
@@ -1,21 +1,22 @@
...
-	rep(i,0,n) if (f.count(b = b * a % m))
-		res = min(res, f[b] * n - i - 1);
-	return res;
+    ll n = (ll) sqrt(m) + 1, e = 1, f = 1, j = 1;
+    unordered_map<ll, ll> A;
+    while (j <= n && (e = f = e * a % m) != b % m)
+        A[e * b % m] = j++;
+    if (e == b % m) return j;
+    if (__gcd(m, e) == __gcd(m, b)) 
+        rep(i, 2, n + 2) if (A.count(e = e * f % m))
```
simonlindholm, line 19: nit: `rep(i,2,n+2)` for consistency

bjorn-martinsson: Fixed

PR #176, opened 2020-06-05
Let `eulerWalk()` return the edge indices by default

Mohammad-Yasser:
I think that you already have thought about this and have your reasons why you didn't do that.
If you'd like to point out the reasons, I'll be grateful, because I thought of possible reasons and couldn't find any.
The only disadvantage is that the code is a bit longer in my opinion.

The changes:
- Let the default behavior be to return the edge indices.
Justification: If you don't want them, you'll just not use them, but if you want them, you'll have to do a change (minor yet still a change).
- Let the default `src` be a non-isolated node.
Justification: If I want to get a cycle and the graph can have isolated nodes, I'll need to specify the node myself.
Bonus: What about searching for an odd-degree node first to account for the case of Euler Path? I think this should be done too.
- Don't create a new vector just to return.
Justification: It's a nitpick, I'm not sure if it's worthy.

### DIFF
--- a/content/graph/EulerWalk.h
+++ b/content/graph/EulerWalk.h
@@ -6,26 +6,44 @@
  * Description: Eulerian undirected/directed path/cycle algorithm.
  * Input should be a vector of (dest, global edge index), where
  * for undirected graphs, forward/backward edges have the same index.
- * Returns a list of nodes in the Eulerian path/cycle with src at both start and end, or
- * empty list if no cycle/path exists.
- * To get edge indices back, add .second to s and ret.
+ * Returns a list of pairs of the nodes in the Eulerian path/cycle with src at both start and end,
+ * and the edge indices, or empty list if no cycle/path exists.
  * Time: O(V + E)
  * Status: stress-tested
  */
 #pragma once
 
-vi eulerWalk(vector<vector<pii>>& gr, int nedges, int src=0) {
-	int n = sz(gr);
-	vi D(n), its(n), eu(nedges), ret, s = {src};
-	D[src]++; // to allow Euler paths, not just cycles
-	while (!s.empty()) {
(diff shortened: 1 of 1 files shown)


simonlindholm:
I'm not totally opposed to having it return the edge indices, since modifying code can indeed be tricky. The main reason I haven't done it is that it complicates both code and description/usage.

As is I'm not a fan of this PR -- it bloats code size *a lot*, and KACTL cares deeply about that. This is also why there's no auto-detection of where to start (putting "set src to a non-isolated node, with odd degree if possible" in the description would be better IMO), and why it uses {ret.rbegin(), ret.rend()}. Codegolfing it back into its original form would help a bit, not sure if worth taking still though. (Admittedly the original code is a bit over-golfed with multiple statements on one line, which KACTL usually doesn't do... but I feel like it's readable enough.)

Mohammad-Yasser:
Makes sense yeah.

I have another suggestion that maybe like a middle ground, introduce two helper functions:
- `vector<pii> eulerPath(graph, nedges)`: Finds an odd-degree node, if any, or a positive-degree node, and call `eulerWalk` with it as `src`.
- `vector<pii> eulerCycle(graph, nedges)`: Finds a positive-degree node and call `eulerWalk` with it as `src`, and return `{}` if the output is not a cycle.

That way the old function will remain as it is, but it'll be optional to use these two functions.
I'm assuming that you don't want much code in this function, but don't mind to have other non-essential functions. Ofc if there should not be non-essential code at all, then this suggestion won't be suitable as well.

simonlindholm:
That's a nice thought, but unfortunately doesn't help with the goal of code size -- KACTL tries to keep to an ICPC-imposed limit of 25 pages.

Chillee:
You can also just make it a separate file, and include `eulerWalk.h`. That way, those that want it can include it, and others don't need to.

PR #177, opened 2020-08-30
Make `LineContainer.inf` static

Mohammad-Yasser:
This allows us to use the assignment operator `=` with `LineContainer` objects.
Without making it static, trying to do `line_container1 = line_container2;` results in a compilation error.


simonlindholm (approved):
Makes sense

PR #179, opened 2020-09-09
Use relative link to refer files

finn0:
It's always a good idea to prefer relative links to refer files and
folders over their absolute link. This way, we can ensure people are
referring to correct file or folder, even if someone else has forked the
project.


simonlindholm:
Nice! Didn't know this syntax worked.

PR #181, opened 2020-10-06
Remove warnings from structured bindings by upgrading to -std=c++17

BarrensZeppelin:


simonlindholm:
It's not totally clear that we should do this since some judges are still stuck on C++14 (and some even on C++11 and old enough GCC versions that structured bindings are an error instead of a warning, so we've tried to keep away from structured bindings when the code size win isn't too large). I know @Chillee has opinions on this.

BarrensZeppelin:
Maybe the change should only be performed for the stress tests initially, then. :slightly_smiling_face: 
A separate issue can be used to discuss if moving to C++17 is a good idea overall.

simonlindholm:
Yeah, doing so for the stress tests seems sensible, seeing as we run a separate C++14 test-compile on the code in CI anyway.

simonlindholm:
Thanks!

Chillee:
For example, judges like USACO (the primary high school olympiad in the US) as well as many ICPC regionals still use C++11 with gcc 4.8.

PR #182, opened 2020-10-07
Fix bug in LineHullIntersection.

Xylofo:
lineHull should take poly by reference to avoid copying. This improves runtime from linear to logarithmic.


simonlindholm:
Oops!

PR #187, opened 2021-01-08
fix wrong cut in global min cut

Veladus:
Currently the cut returned by `GlobalMinCut.h` doesn't achieve the value of the minimal cut. This is also noted on the [Wikipedia](https://en.wikipedia.org/wiki/Stoer%E2%80%93Wagner_algorithm) page, which uses the same code - upto coding conventions. This PR fixes this problem.

### DIFF
--- a/content/graph/GlobalMinCut.h
+++ b/content/graph/GlobalMinCut.h
@@ -10,7 +10,9 @@
 
 pair<int, vi> getMinCut(vector<vi>& weights) {
 	int N = sz(weights);
-	vi used(N), cut, best_cut;
+	vi used(N), best_cut;
+	vector<vi> cuts(N);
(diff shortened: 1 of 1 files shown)


simonlindholm:
Ouch. Really should have removed this from the repo for being useless and lacking testing... The implementation isn't even very well written. I guess I'll do penance by writing a test, and hopefully I can find time to rewrite it from scratch later.

simonlindholm:
Added a test in 8af10e15a5dcbf26ca75a76194a16fcbda94909c. Thanks a lot for the PR!

simonlindholm:
... and cleaned up in 3d44557b4dc8b9708e8eeb6056bb20e25e468b85.

PR #189, opened 2021-02-25
Use c++17 in test-compiles.sh

Tyilo:


simonlindholm:
As long as content/contest/.bashrc uses -std=c++14 I don't think we should do this. See also #181.

PR #192, opened 2021-03-27
Zfunc: string => const string&

blin00:


simonlindholm:
Sure, I guess this is fine. A 5% speedup apparently, at the cost of 7 chars, though it's hard to see cases where this would be in an inner loop and actually matter perf-wise. Sorry for extremely slow merge.

PR #202, opened 2021-10-11
Add O(N*max(weight)) solution to Knapsack problem

Gullesnuffs:


bjorn-martinsson:
Nice implementation! About "golfing it", I think something like this is shorter (and arguably nicer)
```cpp
    vi u, v(m*2+1, -1);
    u[a+m-t] = b+1;
    rep(i,b,sz(w)) {
        u = v;
        ...
    }
``` 
I'd even argue that doing it like this is natural even if you weren't golfing it, since you can think of `u` as `v` without any changes.

bjorn-martinsson:
I'm very confused about 

https://github.com/kth-competitive-programming/kactl/blob/648db472abd1987c159862aa5582c19082723399/content/various/FastKnapsack.h#L18

Why is it `b + 1` and not `b`? Had `b` been the index to the last element in the "break solution", then `b + 1` would make sense. But `b` is the size of the break solution, and not the index.

PR #207, opened 2021-12-19
Better compiler flags during `make test-compile`

lrvideckis:
https://codeforces.com/blog/entry/15547: better compiler flags during test compile.

Backstory on how I found this: I stole your build+test scrips to better test my CP repo https://github.com/lrvideckis/Programming-Team-Code/tree/master/scripts (great scripts BTW!). And I found a bunch of compiler warnings (one warning was due to a bug!). 

Also what do you think about using `-std=c++2a`?

### DIFF
--- a/doc/scripts/test-compiles.sh
+++ b/doc/scripts/test-compiles.sh
@@ -4,4 +4,5 @@ echo "
 #include \"../content/contest/template.cpp\"
 #include \"../$FILE\"
 " >build/temp.cpp
-g++ -Wall -Wextra -Wfatal-errors -Wconversion -std=c++14 build/temp.cpp && rm a.out build/temp.cpp
+# Catching silly mistakes with GCC: https://codeforces.com/blog/entry/15547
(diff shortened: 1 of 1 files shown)


simonlindholm:
Adding `-pedantic -Wno-unused-result -Wshadow -Wformat=2 -Wfloat-equal -Wcast-qual -Wcast-align` seems fine if they don't add any warnings when running on the repo, though I'd expect some false positives.

`-D_GLIBCXX_DEBUG -D_GLIBCXX_DEBUG_PEDANTIC -fsanitize=address -fsanitize=undefined -fno-sanitize-recover=all -fstack-protector` aren't helpful here -- this script only checks that the files compile and then immediately removes the artifacts, and adding runtime error checks just worsens compile time (the lack of -g is similar).

`-DLOCAL -std=c++2a` aren't used by KACTL's real compile command (https://github.com/kth-competitive-programming/kactl/blob/main/content/contest/.bashrc) so we shouldn't use them here. Updating -std= in particular has come up in the past, but some judges are still stuck on old GCC versions, see #189. We can reevaluate in a couple of years if the need comes up.

lrvideckis:
I went through the codeforces blog in more detail, and added some missing flags since the last commit (`-Wlogical-op -Wshift-overflow=2 -Wduplicated-cond`).

Good catch with runtime flags. Thoughts on adding `-O2`? From the blog:

> some warnings are only enabled together with optimization

Finally regarding

> seems fine if they don't add any warnings

So they added a lot of warnings: [with -Wshadow](https://pastebin.com/EL6THGBa), [without -Wshadow](https://pastebin.com/Q34KXir0)

PR #210, opened 2022-02-06
removing slower bipartite matcher

lrvideckis:


simonlindholm:
The slow bipartite matching is kept intentionally: https://github.com/kth-competitive-programming/kactl/issues/75#issuecomment-486739832

lrvideckis:
Thoughts on switching DFSMatching to something like this? Your implementation TLE's on https://judge.yosupo.jp/problem/bipartitematching whereas the below implementation doesn't (although both are quadratic).

Also the below implementation is roughtly as short which was one of your concerns

https://codeforces.com/blog/entry/58048?#comment-417398

lrvideckis:
If you think it's a good idea, I can make another PR

PR #211, opened 2022-04-19
Update WeightedMatching.h

HalfVoxel:
Add implicit requirement. Using it with N > M causes inf loops.


content/graph/WeightedMatching.h
```diff
@@ -8,7 +8,7 @@
  * nodes are in two matchings and the sum of the edge weights is minimal. Takes
  * cost[N][M], where cost[i][j] = cost for L[i] to be matched with R[j] and
  * returns (min cost, match), where L[i] is matched with
- * R[match[i]]. Negate costs for max cost.
+ * R[match[i]]. Negate costs for max cost. Requires N <= M.
```
simonlindholm, line 11:
```suggestion
 * R[match[i]]. Negate costs for max cost. Requires $N \le M$.
```

simonlindholm (approved):

PR #213, opened 2022-08-25
Change `string& s` to `const string& s`

huynhtrankhanh:
This makes the hashing code easier to use. For example, I can write `hashString("abc")` directly instead of having to explicitly declare a new string.


simonlindholm:
I'm fine with this for Hashing-codeforces, but for Hashing which is meant to be typed by hand this just adds extra characters that are seldom needed, and it would add another line to the pdf too by making `hashString` overrun the 63-char line limit. In the case where you get compile errors I think it's easy enough to go back and add the `const` manually.

(Speaking of Hashing-codeforces we should really reconsider its existence now that CF has added 64-bit support, maybe replace it by just the randomization code... For another issue though.)

PR #218, opened 2022-11-23
Fixed signs in one of the recurrences' expressions

RobotoLev:
Seems like minuses should be there, as initial recurrence uses previous values of `a` on the right side of the equation.

Source: Combinatorics and Number Theory lection.

I didn't recompile pdf with this improvement.


simonlindholm:
Ooops. Good catch!

PR #222, opened 2023-03-21
Divide & Conquer DP

Vidit-Jain:
Replaced the kactl version of the template with one similar to that on CP-Algorithms as it's arguably easier and more intuitive to use.
Verify if the template follows good standards for the notebook.


Vidit-Jain:
Apologies, made pull request on the wrong repository

PR #228, opened 2023-07-19
Modularize lazy segment tree

### LINKED ISSUE #216: Beginner-friendliness of Lazy Segment Tree
Compared to the Lazy Segment Tree, the Segment Tree template is very beginner friendly.
The three lines you have to change to alter the behaviour of the tree are adjacent, and you don't have to understand the rest of the code to make those changes.

The Lazy Segment Tree template does not have this nice property.
Making _that_ tree compute e.g. sums instead of maximums requires tricky changes in multiple places.
If you know how these trees work you might be able to do it, but otherwise it's basically impossible.
It's probably not possible to make it as nice to modify the behaviour as it is in the ordinary Segment Tree, but a little help can go a long way. :slightly_smiling_face: 

I don't have a concrete idea for how to improve it (which is why I'm submitting an issue instead of a pull request), but maybe we can discuss.
I know that beginner-friendliness is not a primary goal of `kactl`, but maybe it's possible to improve it without sacrificing the other goals.

### LINKED ISSUE #227: Lazy segment tree is difficult to modify for other operations
The lazy segment tree of this is repo very useful when exactly the operations that it has are needed, but when other are needed it is difficult to modify because it is not very clear where the lazy values are propagated and where the operations are made.
In normal segment tree on the other hand is very easy to modify the operation because it is in a separate function.
I propose making a similar separation for lazy segment tree.
Probably it wold look something like this: https://github.com/IvanRenison/icpc-team-notebook-el-vasito/blob/update_STree_lazy/data_structures/segment_tree_lazy.cpp (note that that lazy segment tree is different because it is not lazy in memory allocation)

If you are interested in this improvement I can make it (including the tests) and send a pull request.

IvanRenison:
resolves #227, resolves #216

I have separated the operations of lazy segment tree in to functions so now it should be easy to modify.
I tested a modified operation with [CSES Range Updates and Sums](https://cses.fi/problemset/task/1735/): https://cses.fi/paste/7b5c0925b9926d5f640805/

### DIFF
--- a/content/data-structures/LazySegmentTree.h
+++ b/content/data-structures/LazySegmentTree.h
@@ -16,51 +16,54 @@
 
 const int inf = 1e9;
 struct Node {
+	typedef int T; // data type
+	struct L { int mset, madd; }; // lazy type
+	const T tneut = -inf;     // neutral elements
+	const L lneut = {inf, 0};
+	T f (T a, T b) { return max(a, b); } // (any associative fn)
+	T apply (T a, L b) {
+		return b.mset != inf ? b.mset + b.madd : a + b.madd;
+	} // Apply lazy
+	L comb(L a, L b) {
+		if (b.mset != inf) return b;
+		return {a.mset, a.madd + b.madd};
+	} // Combine lazy
+
 	Node *l = 0, *r = 0;
-	int lo, hi, mset = inf, madd = 0, val = -inf;
-	Node(int lo,int hi):lo(lo),hi(hi){} // Large interval of -inf
-	Node(vi& v, int lo, int hi) : lo(lo), hi(hi) {
+	int lo, hi; T val = tneut; L lazy = lneut;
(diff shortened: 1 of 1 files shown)


IvanRenison:
Is this pull request going to be accepted or closed?

IvanRenison:
@simonlindholm

simonlindholm:
> Is this pull request going to be accepted or closed?

Sorry for failing to respond to this. I've been hesitant to generalize the lazy segtree is the past, because there's so many possible variations, e.g. the classical sum/min/max interacting with add-on-range+set-on-range, but also wanting to support complex ad hoc pushing conditions, and I've felt like any generalization risks making things abstract and hard to understand and modify. The lazy segtree is due for a rewrite though, e.g. due to its bad performance, and there's been multiple people asking for this... So maybe. Although, KACTL is kinda in low maintenance mode atm, with me no longer actively competing, so I reckon it will take a while before anything happens on this front.

PR #229, opened 2023-07-31
golf z func

lrvideckis:
The idea behind this change is:

> each iteration of the while loop will increase the right border  
$r$  of the match segment

https://cp-algorithms.com/string/z-function.html#asymptotic-behavior-of-the-algorithm

So basically if we successfully enter in the while loop, then we know the condition `(i + z[i] > r)` is true

here's an AC to show it's still O(n) https://codeforces.com/contest/126/submission/216604260

### DIFF
--- a/content/strings/Zfunc.h
+++ b/content/strings/Zfunc.h
@@ -13,9 +13,7 @@ vi Z(const string& S) {
 	rep(i,1,sz(S)) {
 		z[i] = i >= r ? 0 : min(r - i, z[i - l]);
 		while (i + z[i] < sz(S) && S[i + z[i]] == S[z[i]])
-			z[i]++;
-		if (i + z[i] > r)
-			l = i, r = i + z[i];
+			z[i]++, l = i, r = i + z[i];
 	}
 	return z;
 }


lrvideckis:
btw, the same idea can be used here

https://github.com/kth-competitive-programming/kactl/blob/main/content/strings/Manacher.h#L19-L20

bjorn-martinsson:
On the subject on codegolfing the z function. You can save one more character by changing 
```cpp
z[i] = i >= r ? 0 : min(r - i, z[i - l]);
``` 
into
```cpp
z[i] = i > r ? 0 : min(r - i, z[i - l]);
```
This works since `z` is non-negative, so removing the `=` doesn't make any difference.

lrvideckis:
lib checker AC

https://judge.yosupo.jp/submission/153324

lrvideckis:
actually, this way is even more golfed

https://judge.yosupo.jp/submission/153326

lrvideckis:
https://judge.yosupo.jp/submission/153328

IvanRenison:
@lrvideckis I tested this version o Zfunc and the current one, and in the [problem](https://codeforces.com/gym/105053/problem/H) in witch I tested it it was faster with the current one that with this one. Did you made a more exhaustive test about speed? (I only tested it in one problem, so it may not be something general)

bjorn-martinsson:
> @lrvideckis I tested this version o Zfunc and the current one, and in the [problem](https://codeforces.com/gym/105053/problem/H) in witch I tested it it was faster with the current one than with this one. Did you made a more exhaustive test about speed? (I only tested it in one problem, so it may not be something general)

Could you link to your submissions comparing the two?

IvanRenison:
> Could you link to your submissions comparing the two?

Current versión: https://codeforces.com/gym/105053/submission/254710440
Your versión: https://codeforces.com/gym/105053/submission/254709569

Files in github:
Current versión: https://github.com/IvanRenison/ProgrammingProblems/blob/e73f34d3d818d37ea0c341e976578f70cc5e89b3/ICPC%20contests/Latin%20American%20Super-Regional%202024/H.cpp
Your versión: https://github.com/IvanRenison/ProgrammingProblems/blob/152993eb7982d2585833ff5d574654601bd6a69e/ICPC%20contests/Latin%20American%20Super-Regional%202024/H.cpp

PR #233, opened 2023-10-25
Use tabs for indentation in all tests

IvanRenison:
Some tests had spaces in the indentation, in these pull request I changed them all be tabs, so that the tests have the same style that the code

### DIFF
--- a/stress-tests/data-structures/LineContainer.cpp
+++ b/stress-tests/data-structures/LineContainer.cpp
@@ -6,38 +6,38 @@ namespace other {
 // source: https://github.com/niklasb/contest-algos/blob/master/convex_hull/dynamic.cpp
 const ll is_query = -(1LL<<62);
 struct Line {
-    ll m, b;
-    mutable function<const Line*()> succ;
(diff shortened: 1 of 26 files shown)


stress-tests/data-structures/LineContainer.cpp
```diff
@@ -6,38 +6,38 @@ namespace other {
...
-    ll m, b;
-    mutable function<const Line*()> succ;
-    bool operator<(const Line& rhs) const {
-        if (rhs.b != is_query) return m < rhs.m;
-        const Line* s = succ();
-        if (!s) return 0;
-        ll x = rhs.m;
-        return b - s->b < (s->m - m) * x;
-    }
+		ll m, b;
```
simonlindholm, line 9: This and some other places got indented with two tabs instead of just one, somehow.

IvanRenison: Ok, I will check and fix

IvanRenison:
Fixed

Issue #217, opened 2022-11-15
Wrong Time Complexity for MinCostMaxFlow


bqi343:
https://github.com/kth-competitive-programming/kactl/blob/f6a97d9e5766c90ee71a97a90cd3faf6cff0ca72/content/graph/MinCostMaxFlow.h#L9

As per https://codeforces.com/blog/entry/70740 the runtime should actually be $O(F\cdot E\log V)$, where $F$ is the number of units of flow (plus the time for setpi).

simonlindholm:
Fixed by #237.

PR #237, opened 2023-12-23
Improve MCMF

Matistjati:
This PR improves 3 things:

1. Fixes the time complexity description, as mentioned in #217 
2. No longer uses quadratic memory (adjacency list instead of matrix). Quadratic memory was problematic on problems with sparse graphs such as [tourist](https://open.kattis.com/problems/tourist).
3. Allows parallell/double edges.

Empirically, it also seems to improve performance (see "improved MCMF" [here](https://docs.google.com/spreadsheets/d/11ucYantzFPazqAlAa3J3-wC3NPmcXl3OMAqFhnXPMr8/edit?usp=sharing))

Potential downsides of new implementation: 
1. Loops (edges from a node to itself) don't work (when would you ever want that, though?)
2. Uses pointers to easily reconstruct path, making some parts less readable.

It passes all stress tests and doesn't get any WA's on the kattis problems I've tried. It might be worth to consider using SPFA for setpi if it does not complicate the code too much.

Side note: does anyone know why we can skip using setpi if input costs are positive?

### DIFF
--- a/content/graph/MinCostMaxFlow.h
+++ b/content/graph/MinCostMaxFlow.h
@@ -2,11 +2,11 @@
  * Author: Stanford
  * Date: Unknown
  * Source: Stanford Notebook
- * Description: Min-cost max-flow. Double edges are allowed.
+ * Description: Min-cost max-flow.
  *  If costs can be negative, call setpi before maxflow, but note that negative cost cycles are not supported.
  *  To obtain the actual flow, look at positive values only.
  * Status: Tested on kattis:mincostmaxflow, stress-tested against another implementation
- * Time: $O(F \cdot Elog(V))$ if all costs are positive, where F is max flow. Otherwise $O(F \cdot Elog(V)+VE)$
+ * Time: $O(F \cdot E \log(V))$ where F is max flow. $O(VE)$ for setpi.
  */
 #pragma once
 
@@ -15,8 +15,7 @@
 const ll INF = numeric_limits<ll>::max() / 4;
 
 struct MCMF {
-	struct edge
-	{
+	struct edge {
 		int from, to, rev;
 		ll cap, cost, flow;
 	};
--- a/content/graph/MinCostMaxFlow.h
+++ b/content/graph/MinCostMaxFlow.h
@@ -46,8 +46,7 @@ struct MCMF {
 		while (!q.empty()) {
 			s = q.top().second; q.pop();
 			seen[s] = 1; di = dist[s] + pi[s];
-			for (edge& e : ed[s]) if (!seen[e.to])
-			{
+			for (edge& e : ed[s]) if (!seen[e.to]) {
 				ll val = di - pi[e.to] + e.cost;
 				if (e.cap - e.flow > 0 && val < dist[e.to]) {
 					dist[e.to] = val;
--- a/content/graph/MinCostMaxFlow.h
+++ b/content/graph/MinCostMaxFlow.h
@@ -29,8 +29,8 @@ struct MCMF {
 
 	void addEdge(int from, int to, ll cap, ll cost) {
 		if (from == to) return;
-		ed[from].emplace_back(edge{ from,to,sz(ed[to]),cap,cost,0});
-		ed[to].emplace_back(edge{ to,from,sz(ed[from])-1,0,-cost,0});
+		ed[from].push_back(edge{ from,to,sz(ed[to]),cap,cost,0 });
+		ed[to].push_back(edge{ to,from,sz(ed[from]) - 1,0,-cost,0 });
 	}
 
 	void path(int s) {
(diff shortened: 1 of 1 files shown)


content/graph/MinCostMaxFlow.h
```diff
@@ -2,36 +2,36 @@
  * Author: Stanford
  * Date: Unknown
  * Source: Stanford Notebook
- * Description: Min-cost max-flow. cap[i][j] != cap[j][i] is allowed; double edges are not.
+ * Description: Min-cost max-flow. Double edges are allowed.
```
simonlindholm, line 5:
```suggestion
 * Description: Min-cost max-flow.
```

simonlindholm (commented):
Nice to hear this improves perf, I was worried about that before but I guess it makes sense. (Maybe it would another thing if the vector<vector<>> was a packed matrix.)

I agree self loops aren't interesting -- they could be if they had negative weight but we don't handle negative weight cycles anyway. And them not showing up in the list of edges doesn't feel like a problem at all since you'll be filtering out the subset of edges with flow > 0 anyway if you ever go through it.

content/graph/MinCostMaxFlow.h
```diff
@@ -2,36 +2,36 @@
...
+	MCMF(int N) : N(N), ed(N), seen(N), dist(N), pi(N), par(N) {}
 
 	void addEdge(int from, int to, ll cap, ll cost) {
-		this->cap[from][to] = cap;
-		this->cost[from][to] = cost;
-		ed[from].push_back(to);
-		red[to].push_back(from);
+		if (from == to) return;
+		ed[from].emplace_back(edge{ from,to,sz(ed[to]),cap,cost,0});
+		ed[to].emplace_back(edge{ to,from,sz(ed[from])-1,0,-cost,0});
```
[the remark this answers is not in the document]
Matistjati: Fixed

content/graph/MinCostMaxFlow.h
```diff
@@ -68,15 +63,17 @@ struct MCMF {
 		ll totflow = 0, totcost = 0;
 		while (path(s), seen[t]) {
 			ll fl = INF;
-			for (int p,r,x = t; tie(p,r) = par[x], x != s; x = p)
-				fl = min(fl, r ? cap[p][x] - flow[p][x] : flow[x][p]);
+			for (int p, x = t; par[x] && (p = par[x]->from, x != s); x = p)
```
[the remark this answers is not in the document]
Matistjati: I chose to use python-style ifs to fix the "push into priority queue" from overflowing. A bit stylistically inconsistent, but probably fine?
https://github.com/kth-competitive-programming/kactl/blob/48ca736f0b7ad17d7e1c38fd584d61a81715ab22/content/graph/MinCostMaxFlow.h#L53-L56

After fiddling with the loops, I found a pretty clean solution in the end.
https://github.com/kth-competitive-programming/kactl/blob/48ca736f0b7ad17d7e1c38fd584d61a81715ab22/content/graph/MinCostMaxFlow.h#L67-L74

content/graph/MinCostMaxFlow.h
```diff
@@ -2,36 +2,35 @@
  * Author: Stanford
  * Date: Unknown
  * Source: Stanford Notebook
- * Description: Min-cost max-flow. cap[i][j] != cap[j][i] is allowed; double edges are not.
+ * Description: Min-cost max-flow.
  *  If costs can be negative, call setpi before maxflow, but note that negative cost cycles are not supported.
  *  To obtain the actual flow, look at positive values only.
  * Status: Tested on kattis:mincostmaxflow, stress-tested against another implementation
- * Time: Approximately O(E^2)
+ * Time: $O(F \cdot E \log(V))$ where F is max flow. $O(VE)$ for setpi.
```
simonlindholm, line 9:
```suggestion
 * Time: $O(F E \log(V))$ where F is max flow. $O(VE)$ for setpi.
```
to avoid text overlap in pdf

content/graph/MinCostMaxFlow.h
```diff
@@ -68,26 +64,28 @@ struct MCMF {
...
 	// If some costs can be negative, call this before maxflow:
 	void setpi(int s) { // (otherwise, leave this out)
 		fill(all(pi), INF); pi[s] = 0;
 		int it = N, ch = 1; ll v;
 		while (ch-- && it--)
-			rep(i,0,N) if (pi[i] != INF)
-				for (int to : ed[i]) if (cap[i][to])
-					if ((v = pi[i] + cost[i][to]) < pi[to])
-						pi[to] = v, ch = 1;
+			rep(i, 0, N) if (pi[i] != INF)
```
simonlindholm, line 85:
```suggestion
			rep(i,0,N) if (pi[i] != INF)
```

content/graph/MinCostMaxFlow.h
```diff
@@ -68,26 +64,28 @@ struct MCMF {
...
-				if (r) flow[p][x] += fl;
-				else flow[x][p] -= fl;
+			for (edge* x = par[t]; x; x = par[x->from]) {
+				x->flow += fl;
+				ed[x->to][x->rev].flow -= fl;
+			}
 		}
-		rep(i,0,N) rep(j,0,N) totcost += cost[i][j] * flow[i][j];
-		return {totflow, totcost};
+		rep(i,0,N) for(edge& e : ed[i]) totcost += e.cost*e.flow;
```
simonlindholm, line 76:
```suggestion
		rep(i,0,N) for(edge& e : ed[i]) totcost += e.cost * e.flow;
```

simonlindholm (approved):

PR #240, opened 2024-01-09
Add clarification on gss description

IvanRenison:
Add the text "and no local maximum" in the description of golden section search (gss) to note that gss may not work well if the function has a local maximum in the interval


simonlindholm:
I guess it depends on whether you could endpoints as local minima, but sure.

simonlindholm:
I guess a downside of the new wording and the implication that endpoints aren't to be counted (as they would be local maxima otherwise) is that it implies that monotonic functions aren't allowed, but they are.

PR #247, opened 2024-03-02
Use `ll` instead of `int` everywhere

IvanRenison:


IvanRenison:
Sorry, it was not for here

PR #258, opened 2024-05-03
library checker tests

lrvideckis:
initial PR just to test that I set it up correctly. I plan to add more tests later

### DIFF
--- a/.github/workflows/ccpp.yml
+++ b/.github/workflows/ccpp.yml
@@ -22,3 +22,29 @@ jobs:
       run: make test-compiles
     - name: Run stress tests
       run: make test
+
+  verify:
+    runs-on: ubuntu-latest
+
+    steps:
+    - uses: actions/checkout@v3
+
+    - name: Set up Python
+      uses: actions/setup-python@v4
+
+    - name: Install dependencies
+      run: pip3 install -U online-judge-verify-helper
+
+    # required by cargo-udeps
+    - name: Set up Rust (nightly)
+      run: |
+        rustup set profile minimal
+        rustup install nightly
+        rustup override set nightly
+
+    - name: Run tests
+      env:
+        GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        YUKICODER_TOKEN: ${{ secrets.YUKICODER_TOKEN }}
+        GH_PAT: ${{ secrets.GH_PAT }}
+      run: oj-verify all
--- a/.verify-helper/config.toml
+++ b/.verify-helper/config.toml
@@ -1,0 +1,14 @@
+[[languages.cpp.environments]]
+CXX = "g++"
+CXXFLAGS = [
+	"-O2",
+	"-std=c++17",
+	"-fsanitize=address,undefined",
+	"-fno-sanitize-recover=all",
+	"-fstack-protector",
+	"-D_GLIBCXX_DEBUG",
+	"-D_GLIBCXX_SANITIZE_VECTOR",
+	"-D_GLIBCXX_DEBUG_PEDANTIC",
+	"-D_GLIBCXX_ASSERTIONS",
+	"-D_FORTIFY_SOURCE=2",
+]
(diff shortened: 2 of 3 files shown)


lrvideckis:
can you run the workflow

simonlindholm:
Sure. Do I need to set GITHUB_TOKEN/YUKICODER_TOKEN/GH_PAT to something specific (or can they be removed)? I think we might want a stress-tests/librarychecker subdirectory to make the delineation clear.

lrvideckis:
oh good question, hmm, thb I just followed the steps here

https://online-judge-tools.github.io/verification-helper/installer.html

which generated the .yml file

I know that after the tests run, an automated commit is pushed to the repo (test results are stored in a file) so probably the github tokens are needed

lrvideckis:
no you dont need to set them to anything

lrvideckis:
also looks like I broke the other test script; ill move the tests to ~/library-checker e.g. not in stress-tests/...

should fix it

https://github.com/kth-competitive-programming/kactl/blob/main/doc/scripts/run-all.sh#L8

simonlindholm:
> I know that after the tests run, an automated commit is pushed to the repo (test results are stored in a file) so probably the github tokens are needed

Oh. Is there an option to skip that, or to push it to a separate repo? It sounds spammy

lrvideckis:
Hi, so I just tested it on my repo, and you only need `GH_PAT` (ill remove the others)

as far as I know, there's no way to prevent having that file. the tool ([verification helper](https://github.com/online-judge-tools/verification-helper)) that runs the library checker tests is closely integrated with that file. 

for example one annoying thing about verification helper is it [only runs 10-15 tests per commit](https://github.com/online-judge-tools/verification-helper/issues/190) then pushes the results of those tests to that file. (I commented some workarounds). So I think what would happen if you didn't have the file, is only the first 10-15 tests will run, and none others will run

simonlindholm:
That sounds ... not great? Can we do this as a thing that's run manually from time to time rather than running it in CI? I imagine most of the value will be in the initial run-through.

lrvideckis:
Oh I should mention, once it runs all tests initially, then afterwards, it only re-runs the tests for files which changed in that commit

lrvideckis:
but to answer your question, yes, we can set it up where you can run it manually instead of during github CI

lrvideckis:
For my repo, the times where verification-helper is annoying is when I do some code-mod style change that affects many files, and many tests (more than 15) have to be rerun.

lrvideckis:
"the file" looks like this https://github.com/online-judge-tools/verification-helper/blob/master/.verify-helper/timestamps.remote.json

lrvideckis:
oh also library checker tests are generated randomly, so you could argue there is value in re-testing more over time (in github CI)

lrvideckis:
Hi again! let me know if you still want these tests. I'm still willing to write them, but I would like to set it up as part of github CI/actions. I understand your concern about the [test file](https://github.com/online-judge-tools/verification-helper/blob/master/.verify-helper/timestamps.remote.json), but consider the pros:

1. on each new commit, the tests for those changed files are re-run. So the library is continually tested
2. tests are randomly generated, so library gets more tested over time

I can vouch for the quality of these tests ([look how many libraries use it](https://github.com/search?q=online-judge-verify-helper+path%3A.github)). Library checker is [maintained](https://github.com/yosupo06/library-checker-problems/graphs/contributors) by at least 4 codeforces LGM's. Furthermore, specific test cases are crowd-sourced (they already caught one bug in KACTL (in pollard rho))

---
(opinionated)

If it were up to me, in addition to adding `~/library-checker-tests/` folder, I would also go through the tests in `~/old-unit-tests` and `~/stress-tests` and convert all those tests into the library checker framework (via using the [hello world](https://judge.u-aizu.ac.jp/onlinejudge/description.jsp?id=ITP1_1_A) problem as a "dummy problem"). Then I would delete [run-all.sh](https://github.com/kth-competitive-programming/kactl/blob/main/doc/scripts/run-all.sh) (which is written in a [procedural way; not function](https://unix.stackexchange.com/a/169765) as bash is meant to be)

PR #260, opened 2024-05-11
Use `__lg` to make `treeJump` shorter

IvanRenison:
It breaks the case when P is empty, but thats not an useful case

### DIFF
--- a/content/graph/BinaryLifting.h
+++ b/content/graph/BinaryLifting.h
@@ -5,15 +5,14 @@
  * Source: Folklore
  * Description: Calculate power of two jumps in a tree,
  * to support fast upward jumps and LCAs.
- * Assumes the root node points to itself.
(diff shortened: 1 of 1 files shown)


simonlindholm:
I don't care about the case of empty P, because that's not a tree, but I do care about sz(P) == 1. __lg(0) is undefined, right?

IvanRenison:
Oh, yes, I saw in [this stack overflow answer](https://stackoverflow.com/questions/40434664/what-is-std-lg) that it's -1, and in my machine gives -1, but according to this [codeforces blog](https://codeforces.com/blog/entry/92479) you are right.
I will fix it with `sz(P) < 2 ? 0 :1 + __lg(sz(P) - 1)`

lrvideckis:
if sz(P)==1 then __lg(1)=0 and is not UB, so original code would have worked

PR #263, opened 2024-08-25
Fix a bug of MinCostMaxFlow.h

tht2005:
Fix a bug that occur when maxflow function is called multiple times from different source vertices.

### DIFF
--- a/content/graph/MinCostMaxFlow.h
+++ b/content/graph/MinCostMaxFlow.h
@@ -34,6 +34,7 @@ struct MCMF {
 	}
 
 	void path(int s) {
+		fill(all(par), 0);
 		fill(all(seen), 0);
 		fill(all(dist), INF);
 		dist[s] = 0; ll di;


tht2005:
I add tests for this case.

tht2005:
> It's enough to set par[s] = NULL at the start of maxflow, right? Can you modify the stress test to cover this case?

can you review my code

simonlindholm:
NULL -> 0 to save a few chars. For the test, it doesn't actually check that `maxflow` returns the correct result, only that it doesn't crash. Can you check that too? E.g. test it by doing "construct graph. pick random a,b,c,d. compute e = maxflow(a,b), f = maxflow(c,d). set up the same maxflow instance again. compute f' = maxflow(c, d), e' = maxflow(a, b). check that e = e', f = f'."

Matistjati:
Maybe I've just misunderstood this PR, but I don't think we want to support this? I don't think Simon's stated property holds. The usecase this fixes seems to be "call maxflow on multiple different source vertices in the same weakly connected component", which as far as I'm aware is a somewhat nonsensical computation? Additionally, I suspect that you might get issues with having incorrect potentials on some graphs.

PR #264, opened 2024-08-27
Fix undefined behavior in Suffix Array

ssk4988:
Fixes undefined behavior in the Suffix Array implementation caused by copying a value from the end iterator of the string. I was able to get a runtime error on a problem and then get accepted by making this change to the suffix array implementation only.

### DIFF
--- a/content/strings/SuffixArray.h
+++ b/content/strings/SuffixArray.h
@@ -21,8 +21,8 @@ struct SuffixArray {
 	vi sa, lcp;
 	SuffixArray(string& s, int lim=256) { // or basic_string<int>
 		int n = sz(s) + 1, k = 0, a, b;
-		vi x(all(s)+1), y(n), ws(max(n, lim)), rank(n);
-		sa = lcp = y, iota(all(sa), 0);
+		vi x(all(s)), y(n), ws(max(n, lim)), rank(n);
+		x.push_back(0), sa = lcp = y, iota(all(sa), 0);
 		for (int j = 0, p = 0; p < n; j = max(1, j * 2), lim = p) {
 			p = j, iota(all(y), n - j);
 			rep(i,0,n) if (sa[i] >= j) y[p++] = sa[i] - j;


lrvideckis:
wow, ya'll actually are okay with having UB (knowingly) in your library?

that's terrible

ssk4988:
Oops, I accidentally closed this when deleting some branches

simonlindholm:
> wow, ya'll actually are okay with having UB (knowingly) in your library?

Not really, just lacking time for maintenance.

I'm surprised this caused issues in practice, I wonder what compiler/compiler flags were used. `s[s.size()]` is guaranteed to be 0, so it can't have been one that just defined `char* end() { return &(*this)[size()]; }`. The fix is a bit unfortunate because it triggers a reallocation making the code slightly slower. Maybe
```
		vi x(n), y(n), rank(n), ws(max(n, lim));
		sa = lcp = y; rep(i,0,n) sa[i] = i, x[i] = s[i];
```
would be better? It's slightly fewer chars and doesn't have the reallocation. Although, I guess if the allocations ever matter perf-wise there are better things you can do about them. Shrug. I'll just merge.

lrvideckis:
ah I take it back

ssk4988:
I wanted to add that this [submission](https://codeforces.com/gym/102343/submission/278183767) when viewed in coach mode on CodeForces and compared to its previous solution shows that the only significant change is the push_back and it stops a runtime error. I have another submission where the diff is exactly what is in this commit but I am unable to share it for some reason.

> I'm surprised this caused issues in practice, I wonder what compiler/compiler flags were used. `s[s.size()]` is guaranteed to be 0, so it can't have been one that just defined `char* end() { return &(*this)[size()]; }`. The fix is a bit unfortunate because it triggers a reallocation making the code slightly slower. Maybe
> 
> ```
> 		vi x(n), y(n), rank(n), ws(max(n, lim));
> 		sa = lcp = y; rep(i,0,n) sa[i] = i, x[i] = s[i];
> ```
> 
> would be better? It's slightly fewer chars and doesn't have the reallocation. Although, I guess if the allocations ever matter perf-wise there are better things you can do about them. Shrug. I'll just merge.

I hadn't even considered the allocation issue. Your proposed method seems fine.

simonlindholm:
> I wanted to add that this [submission](https://codeforces.com/gym/102343/submission/278183767) when viewed in coach mode on CodeForces and compared to its previous solution shows that the only significant change is the push_back and it stops a runtime error.

The issue with this submission is that `string` was changed to `vector<int>`, and `vector<int>` does not have the behavior of guaranteeing `s[s.size()] == 0` that strings due because of nul termination. (This is the reason for the comment about `basic_string<int>` next to the constructor.) In order to make things work with `vector<int>` you additionally need to change the `for(k && k--, j = sa[rank[i] - 1]; s[i + k] == s[j + k]; k++);` loop not to read out of bounds, or else you're relying on luck.

That said, this is very subtle and I'm sympathetic to the cause of removing that footgun. Maybe we should do `s.push_back(0)` at the start instead? That probably also serves as a good reminder not to put other zeroes in the vector.

user202729:
Randomly come across this.

1. Surprisingly, it is actually undefined behavior: https://en.cppreference.com/w/cpp/string/basic_string/end and MSVC actually checks for this in debug mode: https://github.com/microsoft/STL/issues/4346
2. But cppreference also [says](https://en.cppreference.com/w/cpp/string/basic_string) `*(s.begin() + s.size())` has value `CharT()` (a null terminator)
3. Also surprisingly, `_GLIBCXX_DEBUG` doesn't check for this. The following program runs without error in both g++ and clang++:

	<details>
	```cpp
	#define _GLIBCXX_DEBUG 1
	#define _GLIBCXX_DEBUG_PEDANTIC 1
	#include <bits/stdc++.h>
	using namespace std;
	int main() {
		string a(3, '0');
		cout<<int(*(a.begin()-1))<<'\n';
	}
	```
	</details>

4. It looks like `x(s.data(),s.data()+sz(s)+1)` should be safe. After all `s.data()` is guaranteed to return `char*`.
5. To avoid the bug, I feel the best way is to leave a `// v must be basic_string, not vector` comment there.

simonlindholm:
I pushed 7515986a1dd563e5bf223fe958248dfa15b81f44 to remove the `vector<int>` foot-gun.

PR #279, opened 2024-11-03
Clarify inclusive-exclusive ranges

JJCUBER:


simonlindholm:
I think the ranges on mo on trees are inclusive-inclusive actually! Clarifying this is a good idea.

PR #280, opened 2024-12-01
constant factor speedup (vector -> array)

lrvideckis:


lrvideckis:
problem https://codeforces.com/contest/718/problem/C

current Matrix code: https://codeforces.com/contest/718/submission/294132074 4702 ms 
using array instead of matrix: https://codeforces.com/contest/718/submission/294132337 593 ms

simonlindholm:
Huh, I was sure the compiler would be able to optimize this. Makes sense though.

PR #284, opened 2025-01-23
Fix some issues that may cause catastrophic floating point error

user202729:
Made the following changes

* change `p+d*t` to `q+d*(t-1)`. The problem is if `-a+sqrt(det) ≫ 1`, then the term `arg(v, q)` is supposed to be identically zero, however if `q ≈ (0, 0)` then it's possible that `p+d ≉ q` and `arg(v, q) ≉ 0`, which results in a large error. See `a.cpp` in the attached file below.
* even if `side != (s.cross(e, prev) < 0)` is true, it is possible that the 4 points are almost collinear that `lineInter` returns the second value being `(0, 0)`. With this change, whenever `(a < 0) != (b < 0)` then `0 ≤ a / (a - b) ≤ 1`, which means the point being pushed is always on the line between `prev` and `cur` i.e. no catastrophic error can happen. See `b.cpp` in the attached file below.

Attachment (demonstration of catastrophic failure): [a.zip](https://github.com/user-attachments/files/18520565/a.zip)


simonlindholm:
These are fantastic catches. Thank you so much!

PR #291, opened 2025-01-29
Create pull_request_template.md

tyilo:
Add PR template to try to prevent people from creating PR's by mistake.

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -1,0 +1,6 @@
+<!--
+Make sure you want to submit a pull request to the upstream kactl project.
+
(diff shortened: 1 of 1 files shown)


simonlindholm:
Right. Yeah, it's a reasonable idea. I'm not in love with the suggested wording though, the second sentence makes it sound like the problem is with unwanted changes rather than accidental ones, and the template will apply even to PRs to forks where the first sentence doesn't make sense.

tyilo:
Fair enough. Any suggestions?

simonlindholm:
Not off the top of my head, this kind of wordsmithing always takes me forever.

PR #307, opened 2026-01-03
Pass points by value

Matistjati:
In geometry, it's mixed whether points are passed as `P&`, `const P&` or `P`. We should probably be consistent. Personally, I prefer just P. We should never have `P&`: then you can't do reasonable things such as:
```c++
auto eval = [&](double p) {
    return segDist(a, b, c + delta * p);
};
```

### DIFF
--- a/content/geometry/Point3D.h
+++ b/content/geometry/Point3D.h
@@ -12,19 +12,18 @@
 
 template<class T> struct Point3D {
 	typedef Point3D P;
-	typedef const P& R;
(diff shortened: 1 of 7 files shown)


Matistjati:
We probably shouldn't merge this yet. I haven't tested these changes, and there might be more changes to make. But I want to get the discussion going @simonlindholm

simonlindholm:
Sounds like a good idea to me.

PR #308, opened 2026-01-13
Add shorter and faster Hopcroft-Karp implementation

ahsoltan:
About 2x faster on library checker tests than the old one:
* (old) 1054ms [submission](https://judge.yosupo.jp/submission/345557)
* (new) 585ms [submission](https://judge.yosupo.jp/submission/345556)

Also about 2x shorter.

### DIFF
--- a/content/graph/hopcroftKarp.h
+++ b/content/graph/hopcroftKarp.h
@@ -1,61 +1,34 @@
 /**
- * Author: Chen Xing
- * Date: 2009-10-13
+ * Author: Adam Soltan
+ * Date: 2026-01-13
  * License: CC0
- * Source: N/A
  * Description: Fast bipartite matching algorithm. Graph $g$ should be a list
- * of neighbors of the left partition, and $btoa$ should be a vector full of
- * -1's of the same size as the right partition. Returns the size of
- * the matching. $btoa[i]$ will be the match for vertex $i$ on the right side,
+ * of neighbors of the left partition, and $r$ should be a vector full of
+ * $-1$'s of the same size as the right partition. Returns the size of
+ * the matching. $r[i]$ will be the match for vertex $i$ on the right side,
  * or $-1$ if it's not matched.
- * Usage: vi btoa(m, -1); hopcroftKarp(g, btoa);
- * Time: O(\sqrt{V}E)
- * Status: stress-tested by MinimumVertexCover, and tested on oldkattis.adkbipmatch and SPOJ:MATCHING
+ * Time: O(E \sqrt{V})
+ * Status: stress-tested by MinimumVertexCover and tested on Library Checker
  */
 #pragma once
 
-bool dfs(int a, int L, vector<vi>& g, vi& btoa, vi& A, vi& B) {
-	if (A[a] != L) return 0;
-	A[a] = -1;
-	for (int b : g[a]) if (B[b] == L + 1) {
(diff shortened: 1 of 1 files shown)


simonlindholm:
Hm, as far as I can tell, the BFS here finds nodes at all distances, rather than stopping once the layer with shortest distance is reached. This feels like a bug and probably means that its worst-case performance is worse than O(E sqrt V)?

(Very nice to improve the Hopcroft-Karp! It's a remnant of the olden days and not at all well optimized.)

ahsoltan:
I don't think this impacts complexity, as the O(E*sqrt(V)) complexity comes from the fact that there will be at most sqrt(V) iterations of the outer loop ([source](https://cp-algorithms.com/graph/dinic.html#unit-networks)).
I'll test the early-break optimization on LC once it comes back up, but I don't expect it to be significant.

ahsoltan:
From a few tests ([one](https://judge.yosupo.jp/submission/347641), [two](https://judge.yosupo.jp/submission/347630)) it seems that **not** breaking is significantly faster. Looks like computing all the distances allows a single dfs phase to push more than just the shortest paths. So I guess it should be fine to leave it that way?

simonlindholm:
Hm, yes, upon further reflection and testing it does seem like this is actually correct. One thing I find subtle is the fact that distances are only tracked on the LHS, so the DFS might find a path where the very last node found has another shorter path to it. This invalidates the property that levels don't decrease in the next phase. However, the property *is* maintained for the levels that are less than the shortest distance to an unmatched RHS node, and this is all we need in order to argue that the shortest distance to increase by one in each phase. Stress testing bears this out.

And yeah, it makes total sense that not breaking helps performance in practice. Will merge, thanks for the PR!

content/graph/hopcroftKarp.h
```diff
@@ -1,61 +1,34 @@
...
+	auto dfs = [&](auto f, int u) -> bool {
+		int t = exchange(d[u], 0) + 1;
+		for (int v : g[u])
+			if (r[v] == -1 || (d[r[v]] == t && f(f, r[v])))
+				return l[u] = v, r[v] = u, 1;
+		return 0;
+	};
+	for (int t = 0, f = 0;; t = f = 0, d.assign(n, 0)) {
+		rep(i, 0, n) if (l[i] == -1) q[t++] = i, d[i] = 1;
+		rep(i, 0, t) for (int v : g[q[i]]) {
```
simonlindholm:
```suggestion
		rep(i,0,n) if (l[i] == -1) q[t++] = i, d[i] = 1;
		rep(i,0,t) for (int v : g[q[i]]) {
```

content/graph/hopcroftKarp.h
```diff
@@ -1,61 +1,34 @@
...
+			if (r[v] == -1) f = 1;
+			else if (!d[r[v]]) d[r[v]] = d[q[i]] + 1, q[t++] = r[v];
 		}
-		/// Use DFS to scan for augmenting paths.
-		rep(a,0,sz(g))
-			res += dfs(a, 0, g, btoa, A, B);
+		if (!f) return res;
+		rep(i, 0, n) if (l[i] == -1) res += dfs(dfs, i);
```
simonlindholm:
```suggestion
		rep(i,0,n) if (l[i] == -1) res += dfs(dfs, i);
```

PR #311, opened 2026-03-18
minor golf

lrvideckis:


simonlindholm:
Eh, I don't really care for this, it's the same number of lines and more complex

PR #312, opened 2026-03-27
minor golf

lrvideckis:


simonlindholm:
This requires C++20. I'm not sure all judges use that yet, may be best to wait a few more years still?

PR #314, opened 2026-04-04
golf min vertex cover

lrvideckis:
min vertex cover now depends on hopcroft karp instead of dfs matching

see test file for updated usage

I did change how the cover is returned to essentially 2 vector of ints: I prefer this way becuase it allows you to check if a node is in the cover (if cl[u] == 1)

### DIFF
--- a/content/graph/MinimumVertexCover.h
+++ b/content/graph/MinimumVertexCover.h
@@ -9,25 +9,16 @@
  */
 #pragma once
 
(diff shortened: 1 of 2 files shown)


lrvideckis:
oh ya, it passes https://onlinejudge.u-aizu.ac.jp/problems/GRL_7_A and https://judge.yosupo.jp/problem/bipartitematching

